PR #5, opened 2022-05-03
chore: add ESLint and Prettier lint rules

rowanmanning:
This adds in ESLint and Prettier lint rules and sets up the commands to run them both manually and on commit. The rules themselves were copied from dotcom-debug-user so that we have consistent tooling between projects.


.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+	"semi": true,
+	"singleQuote": true,
+	"bracketSpacing": true,
+	"arrowParens": "always",
+	"trailingComma": "none",
+	"useTabs": true
```
JSRedondo, line 7: nickpick: I fancy adding the tabWidth along with the useTabs property, otherwise...would the code editor take the default one? 🤔

.eslintrc.json
```diff
@@ -0,0 +1,14 @@
+{
+	"extends": ["@financial-times/eslint-config-next"],
+	"parserOptions": {
+		"ecmaVersion": 2020
```
JSRedondo, line 4: nickpick: Just wondering about the es version. Why not 2022? Just

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
JSRedondo, line 2: nickpick: Just curious. How do you come out with this folder name. Is it the husky standard?

JSRedondo (approved):

.prettierrc.json
rowanmanning: Yep it will, but I think we prefer that because everyone has an opinion about how many spaces a tab should make up. If we leave a tab-width definition out of this file then each person can configure their editor however they want 🙂  this is the same for `dotcom-debug-user`

.eslintrc.json
rowanmanning: We're currently supporting Node.js 14 and 16 (because we have apps running on both) and based on https://node.green/ it looks like Node.js 14 doesn't have particularly good ES2021 or ES2022 support.

.husky/pre-commit
rowanmanning: Yep this is automatically added to the start of every bash file that Husky manages and so if we don't include it then Husky won't work correctly

.prettierrc.json
JSRedondo: Yep, maybe I did the same comment for dotcom-debug-user 😄

PR #6, opened 2022-05-03
chore: add unit test runner

rowanmanning:
This adds Jest as a unit test runner. We have it configured to run any
test files within the packages, which it will do in parallel. This means
our tests should remain nice and fast regardless of how many packages we
add to the repo.

I added an example test just to illustrate where they'll be written.
The tests should also automatically run in this PR.

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,0 +1,14 @@
+{
+	"clearMocks": true,
+	"collectCoverage": true,
+	"collectCoverageFrom": [
+		"**/*.js"
+	],
+	"coveragePathIgnorePatterns": [
+		"/node_modules/"
+	],
+	"projects": [
+		"<rootDir>/packages/*"
+	],
+	"testEnvironment": "node"
+}
--- a/jest.config.json
+++ b/jest.config.json
@@ -2,7 +2,8 @@
 	"clearMocks": true,
 	"collectCoverage": true,
 	"collectCoverageFrom": [
-		"**/*.js"
+		"**/*.js",
+		"!**/test/**"
 	],
 	"coveragePathIgnorePatterns": [
 		"/node_modules/"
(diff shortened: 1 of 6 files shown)


jest.config.json
```diff
@@ -0,0 +1,14 @@
+{
+	"clearMocks": true,
+	"collectCoverage": true,
+	"collectCoverageFrom": [
+		"**/*.js"
```
JSRedondo, line 5: suggestion: Maybe we should leave test files out the coverage, adding `"!test/**/*.js"`? Or in the coveragePathIgnorePatterns...

package.json
```diff
@@ -15,16 +15,18 @@
   ],
   "scripts": {
     "lint": "eslint .",
-    "test": "echo 'TODO add jest as a seperate piece of work'",
+    "test": "jest --silent --runInBand",
```
JSRedondo, line 18: nitpick: Just this caught my eye because in the documentation is said the tests run in parallel.
`--runInBand` is safer but also takes longer and makes the tests to run sequentially.

JSRedondo (changes requested):

package.json
rowanmanning: Good spot, I'll remove 👍

jest.config.json
rowanmanning: I can't find the documentation for it at all, but Jest/Istanbul already seems to ignore any files that it uses as part of the tests? I might add this for safety anyway, but it seems fine locally:

Screenshot 2022-05-03 at 11 38 28

rowanmanning: OK I did this because this won't account for things like mocks and fixture files 👍

JSRedondo (approved):

JSRedondo (approved):

PR #9, opened 2022-05-03
feat: initial bootstrapping of the errors package

rowanmanning:
This is a starting point for some central errors. Most of the code in
here is really to act as an example and it will probably change a lot in
future. This PR is mostly here for us to test that the release process
is working as expected.

This will allow me to delete the "example" package (in a separate PR).

### DIFF
--- a/jest.config.json
+++ b/jest.config.json
@@ -8,6 +8,14 @@
 	"coveragePathIgnorePatterns": [
 		"/node_modules/"
 	],
+	"coverageThreshold": {
+		"./packages/**/*.js": {
+			"branches": 100,
+			"functions": 100,
+			"lines": 100,
+			"statements": 100
+		}
+	},
 	"projects": [
 		"<rootDir>/packages/*"
 	],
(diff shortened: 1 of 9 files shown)


jest.config.json
```diff
@@ -8,6 +8,14 @@
 	"coveragePathIgnorePatterns": [
 		"/node_modules/"
 	],
+	"coverageThreshold": {
+		"./packages/**/*.js": {
+			"branches": 100,
+			"functions": 100,
+			"lines": 100,
+			"statements": 100
```
rowanmanning, line 16: I decided what the hell, let's be ambitious and fail the tests if they don't have 100% coverage 😂  if we don't do it from the start then it'll be nearly impossible to do it later

JSRedondo (approved):

jest.config.json
```diff
@@ -8,6 +8,14 @@
 	"coveragePathIgnorePatterns": [
 		"/node_modules/"
 	],
+	"coverageThreshold": {
```
JSRedondo, line 11: Maybe too ambitious at the beginning? 😄

JSRedondo (approved):

jest.config.json
rowanmanning: I've managed it so far 😁

PR #10, opened 2022-05-03
fix: remove the temporary example package

rowanmanning:
Now that we have a real package in #9, we can remove this example package and use `@dotcom-reliability-kit/errors` as an example for future packages.


rowanmanning:
Oh ignore the failing builds, it's because there's no code any more. Once the other PR is merged I'll rebase this and the tests should all pass again.

JSRedondo (approved):

PR #12, opened 2022-05-04
docs: explain our 100% coverage requirement

rowanmanning:
I've added a section to the contributing guide explaining why we expect
100% unit test coverage and how to get around it if the code you're
writing does not need to be covered.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -10,6 +10,7 @@ We're glad you want to contribute to Reliability Kit!
     * [Linters](#linters)
     * [Type safety](#type-safety)
     * [Unit tests](#unit-tests)
+      * [Coverage](#coverage)
   * [Committing](#committing)
     * [Commit type prefixes](#commit-type-prefixes)
     * [Commit linting](#commit-linting)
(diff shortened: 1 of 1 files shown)


docs/contributing.md
```diff
@@ -69,14 +70,31 @@ As with ESLint, we check types in pull requests and errors will block merging, s
...
+```js
+/* istanbul ignore next */
+function example() { console.log('this is not covered'); }
+```
+
+This is better than dropping the required coverage because:
+
+  * We have to make an active and conscious decision to drop a piece of code below 100% coverage
+
+  * The fact that code isn't covered is _documented_ in the code
```
JSRedondo, line 94: Great decision. The issue about the 100% branch coverage is that sometimes it requires additional tests for default parameters.
`function example(a={}, b=true){}`

JSRedondo (approved):

docs/contributing.md
```diff
@@ -69,14 +70,31 @@ As with ESLint, we check types in pull requests and errors will block merging, s
...
 
+#### Coverage
+
+We intentionally fail the unit tests if coverage drops below 100%. This library is meant to help our applications be more reliable and so it's important that we cover as many edge cases as possible. If you have a valid reason for some code not to be covered, e.g. an empty function as a default parameter, then use code comments to disable coverage for that line or block:
+
+```js
+/* istanbul ignore next */
+function example() { console.log('this is not covered'); }
+```
+
```
JSRedondo, line 89: More on this, maybe add this reference to the documentation
[Ignoring code for code coverage](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md)

JSRedondo (approved):

docs/contributing.md
rowanmanning: Good idea 👍

PR #14, opened 2022-05-06
Build errors package

rowanmanning:
This PR is a little large, but I found it difficult to split across multiple PRs because everything is very tightly interlinked. What I have done is written it as [separate smaller commits](https://github.com/Financial-Times/dotcom-reliability-kit/pull/14/commits) which it's possible to review individually. A _lot_ of the added code is in the tests, which maybe don't need to be reviewed quite as thoroughly – I've managed to keep 100% coverage.

## What does this do?

  * I've finished writing the basics for an `OperationalError` class, which allows us to identify when an error is expected vs unexpected. Operational errors also _always_ have an error code so that we can reliably identify the errors when they're logged.

  * I've added an `HttpError` class, which represents a 400–599 response. This is because `http-errors` is one of the most widely-used packages for error handling across our repos – there's clearly a need for something like this.

The best way to understand how these work is to [read the README for the package](https://github.com/Financial-Times/dotcom-reliability-kit/blob/build-errors-package/packages/errors/README.md#dotcom-reliability-kiterrors), as this outlines the full public API for the different error classes.

## Why

I might be jumping the gun a little, but I was having some trouble sleeping and wanted something to work on to engage my brain. Also we've had the proposal up for a week so far and none of the responses have been along the lines of "please don't do this".

## Why not a third party?

I looked into a few third-party libraries for this and opted to not use them under the hood. I'll attempt to explain my thinking here but this is open to argument for sure.

### `OperationalError`

The requirements I think we have for a base error class (which all others extend from) are:

  1. It needs to be marked consistently as "this is an error that we understand". This is so that the rest of our tooling and logging can properly differentiate between known and unknown errors.
  2. It needs to support and encourage having both human-readable and machine-readable identifiers. The error message is for humans to read and might contain information (e.g. `article {ID} is malformed`) which makes it difficult to group the errors because the message may change. The error code is for machines to read and identifies the _class_ of error found and will be the same regardless of slight variations in the error information (e.g. `MALFORMED_ARTICLE`). I think `error.code` would be the best route to go down for this because [it's what Node.js uses](https://nodejs.org/dist/latest-v16.x/docs/api/errors.html#nodejs-error-codes) and means we only have to check one property to determine the class of error.

I looked at the following libraries to see whether they could work for us because they were suggested in the design document (spoiler, I don't think they do):

  * [verror](https://www.npmjs.com/package/verror): this seems to be largely based around strings rather than being able to pass additional data into errors. It also doesn't support setting a `code` or other identifier property easily which I think we need
  * [error-ex](https://www.npmjs.com/package/error-ex): this seems closer to what we want and we could potentially rely on teams extending errors rather than using `code`, but then the rest of our tooling would have to rely on `error.name` or something similar as _well_ as `error.code` (in order to properly identify Node.js built-in errors vs our ones)

### `HttpError`

We do use [http-errors](https://www.npmjs.com/package/http-errors) quite a lot, and I looked into this as an alternative to rolling our own. My main issues were that:

1. This library does not easily support adding a `code` property, so we'd need to extend it somehow anyway. I think it's important that we strive for each error type to have a single (common) property to identify the type of error.
2. Ideally all of our errors will extend a single base error type (in the case of this PR, `OperationalError`). This helps us ensure common required properties are in place and has the benefit of identifying thrown HTTP errors as "known" rather than "unknown"

## What feedback are you looking for?

Anything really. From typos and small suggestions to "your general approach is wrong". Please keep discussion to this PR as it'll be useful in future to have all the conversation in one place if we ever need to explain or justify our approach.

### DIFF
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -5,6 +5,7 @@ A suite of error classes which help you throw the most appropriate error in any 
 
   * [Usage](#usage)
     * [`OperationalError`](#operationalerror)
+      * [`.isErrorMarkedAsOperational`](#operationalerroriserrormarkedasoperational)
   * [Contributing](#contributing)
   * [License](#license)
 
@@ -38,6 +39,17 @@ It works in the same way as a normal error, expecting a message:
 ```js
 throw new OperationalError('example message');
 ```
+
+You can alternatively construct an operational error with a data object. Currently this accepts a `code` property, which must be set to a unique identifier for the type of error which is occurring, and a `message` property which contains a human-readable message:
+
+```js
+throw new OperationalError({
+    message: 'example message',
+    code: 'EXAMPLE_CODE'
+});
+```
+
+Error codes are normalized to be uppercase, alphanumeric, and underscore-delimited.
 
 #### `OperationalError.isErrorMarkedAsOperational()`
 
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -58,13 +58,12 @@ error.message // example message
 error.code // EXAMPLE_CODE
 ```
 
-You may also pass additional properties into an error object, these will be collected and stored on a `data` property on the error. Note: TypeScript will complain about these additional properties, so if you're checking types you will need to ignore the relevant lines:
+You may also pass additional properties into an error object, these will be collected and stored on a `data` property on the error:
 
 ```js
 const error = new OperationalError({
     message: 'example message',
     code: 'EXAMPLE_CODE',
-    // @ts-ignore
     article: 'd92acacb-ac53-4505-aa88-eae4b42de994'
 });
 
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -6,6 +6,8 @@ A suite of error classes which help you throw the most appropriate error in any 
   * [Usage](#usage)
     * [`OperationalError`](#operationalerror)
       * [`.isErrorMarkedAsOperational`](#operationalerroriserrormarkedasoperational)
+    * [`HttpError`](#httperror)
+      * [Why use this over `http-errors`?](#why-use-this-over-http-errors)
   * [Contributing](#contributing)
   * [License](#license)
 
@@ -40,7 +42,7 @@ It works in the same way as a normal error, expecting a message:
 throw new OperationalError('example message');
 ```
 
-You can alternatively construct an operational error with a data object. Currently this accepts a `code` property, which must be set to a unique identifier for the type of error which is occurring, and a `message` property which contains a human-readable message:
+You can alternatively construct an operational error with a data object. This accepts a `code` property, which must be set to a unique identifier for the type of error which is occurring, and a `message` property which contains a human-readable message:
 
 ```js
 throw new OperationalError({
@@ -49,7 +51,12 @@ throw new OperationalError({
 });
 ```
 
-Error codes are normalized to be uppercase, alphanumeric, and underscore-delimited.
+Error codes are normalized to be uppercase, alphanumeric, and underscore-delimited. Error properties can be accessed like any other property:
+
+```js
+error.message // example message
+error.code // EXAMPLE_CODE
+```
 
 #### `OperationalError.isErrorMarkedAsOperational()`
 
@@ -59,6 +66,43 @@ You can test whether an error is operational (known about) either by using the `
 OperationalError.isErrorMarkedAsOperational(new OperationalError('example message')); // true
 OperationalError.isErrorMarkedAsOperational(new Error('example message')); // false
 ```
+
+### `HttpError`
+
+The `HttpError` class extends `OperationalError` and represents an HTTP error status. It can work in the same way as a normal error, expecting a message. In this case it will represent an HTTP `500`:
+
+```js
+throw new HttpError('example message');
+```
+
+You can alternatively construct an HTTP error with a data object. This accepts a `statusCode` property, which is a valid HTTP status code number, as well as all of the properties you can set in [`OperationalError`](#operationalerror):
+
+```js
+throw new HttpError({
+    message: 'your thing was not found',
+    statusCode: 404
+});
+```
+
+It's also possible to create an HTTP error with a status code alone, which will default the message to the corresponding HTTP status message:
+
+```js
+throw new HttpError(404);
+```
+
+Error properties can be accessed like any other property:
+
+```js
+error.message // your thing was not found
+error.statusCode // 404
+error.status // 404
+error.statusMessage // Not Found
+error.code // HTTP_404
+```
+
+#### Why use this over `http-errors`?
+
+The benefit of using this error rather than the excellent [http-errors](https://github.com/jshttp/http-errors#readme) library is that we extend `OperationalError` by default. This means that all HTTP errors you throw are considered "known errors" by the rest of our tooling. We also set a `code` property by default which results in less code in our monitoring dashboards – we don't need to check both `code` and `statusCode` properties to determine the type of error thrown.
 
 
 ## Contributing
--- a/packages/errors/README.md
+++ b/packages/errors/README.md
@@ -58,6 +58,20 @@ error.message // example message
 error.code // EXAMPLE_CODE
 ```
 
+You may also pass additional properties into an error object, these will be collected and stored on a `data` property on the error. Note: TypeScript will complain about these additional properties, so if you're checking types you will need to ignore the relevant lines:
+
+```js
+const error = new OperationalError({
+    message: 'example message',
+    code: 'EXAMPLE_CODE',
+    // @ts-ignore
+    article: 'd92acacb-ac53-4505-aa88-eae4b42de994'
+});
+
+error.data.article // d92acacb-ac53-4505-aa88-eae4b42de994
+```
+
+
 #### `OperationalError.isErrorMarkedAsOperational()`
 
 You can test whether an error is operational (known about) either by using the `isErrorMarkedAsOperational` method. It accepts an error object of any kind and will return `true` if that error has a truthy `isOperational` property and `false` otherwise:
--- a/packages/errors/lib/http-error.js
+++ b/packages/errors/lib/http-error.js
@@ -15,6 +15,16 @@ const OperationalError = require('./operational-error');
 const STATUS_CODES = require('http').STATUS_CODES;
 
 /**
+ * @typedef {Object} HttpErrorData
+ * @property {String} [code]
+ *     A machine-readable error code which identifies the specific type of error.
+ * @property {String} [message]
+ *     A human readable message which describes the error.
+ * @property {Number} [statusCode]
+ *     An HTTP status code.
+ */
+
+/**
  * Class representing an HTTP error.
  */
 class HttpError extends OperationalError {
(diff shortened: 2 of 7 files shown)


packages/errors/lib/operational-error.js
```diff
@@ -44,4 +86,30 @@ module.exports = class OperationalError extends Error {
...
+	}
+}
+
+/**
+ * @typedef {Object} OperationalErrorData
+ * @property {String} [code]
+ *     A machine-readable error code which identifies the specific type of error.
+ * @property {String} [message]
+ *     A human readable message which describes the error.
+ */
```
apaleslimghost, line 113: **thought**: in typescript you could allow arbitrary extra stuff in this object either with an index signature:

```typescript
type OperationalErrorData = {
  code: string
  message: string
  [key: string]: any
}
```

or via an intersection type:

```typescript
type OperationalErrorData = {
  code: string
  message: string
} & Record<string, any>
```

from a quick google i don't know if either of these are possible in JSDoc but might be worth another look, requiring `@ts-ignore` isn't ideal

apaleslimghost: also **suggestion**: it feels weird to me to have this definition at the end of the file? idk why but i think it would flow better with this at the top

packages/errors/README.md
```diff
@@ -37,6 +40,36 @@ It works in the same way as a normal error, expecting a message:
...
+```
+
+You may also pass additional properties into an error object, these will be collected and stored on a `data` property on the error. Note: TypeScript will complain about these additional properties, so if you're checking types you will need to ignore the relevant lines:
+
+```js
+const error = new OperationalError({
+    message: 'example message',
+    code: 'EXAMPLE_CODE',
+    // @ts-ignore
+    article: 'd92acacb-ac53-4505-aa88-eae4b42de994'
```
apaleslimghost, line 66: 👀

packages/errors/README.md
```diff
@@ -46,6 +79,43 @@ OperationalError.isErrorMarkedAsOperational(new OperationalError('example messag
...
+error.statusMessage // Not Found
+error.code // HTTP_404
+```
+
+#### Why use this over `http-errors`?
+
+The benefit of using this error rather than the excellent [http-errors](https://github.com/jshttp/http-errors#readme) library is that we extend `OperationalError` by default. This means that all HTTP errors you throw are considered "known errors" by the rest of our tooling. We also set a `code` property by default which results in less code in our monitoring dashboards – we don't need to check both `code` and `statusCode` properties to determine the type of error thrown.
+
 
 ## Contributing
```
apaleslimghost, line 121: **praise**: this documentation is fantastic!

**suggestion**: i would love to see more fleshed-out, semi-real-world examples of how and when you'd use each of these error classes in context. with the current examples, i could see somebody doing something like:

```js
const response = await fetch('https://example.com')

if(!response.ok) {
	throw new HttpError({
		message: 'error fetching example.com'
		statusCode: response.statusCode
	})
}
```

and calling it a day, which is obviously not a very helpful error to be throwing. examples of how to use these types effectively (and what _not_ to do!) could be really important.

packages/errors/README.md
```diff
@@ -46,6 +79,43 @@ OperationalError.isErrorMarkedAsOperational(new OperationalError('example messag
...
+error.message // your thing was not found
+error.statusCode // 404
+error.status // 404
+error.statusMessage // Not Found
+error.code // HTTP_404
+```
+
+#### Why use this over `http-errors`?
+
+The benefit of using this error rather than the excellent [http-errors](https://github.com/jshttp/http-errors#readme) library is that we extend `OperationalError` by default. This means that all HTTP errors you throw are considered "known errors" by the rest of our tooling. We also set a `code` property by default which results in less code in our monitoring dashboards – we don't need to check both `code` and `statusCode` properties to determine the type of error thrown.
```
apaleslimghost, line 117: **suggestion that would be a lot of work so probably just ignore me but might be worth a thought**: if you were actually using Typescript, `OperationalError` could become an interface, and you could do something like

```ts
import HttpError from 'http-errors'
class OperationalHttpError extends HttpError implements OperationalError {}
```

packages/errors/lib/http-error.js
```diff
@@ -0,0 +1,152 @@
...
+
+		// Default the error message
+		data.message =
+			typeof data.message === 'string'
+				? data.message
+				: HttpError.getMessageForStatusCode(data.statusCode);
+
+		// Message and code are set by the parent class, but we need
+		// to set the status code and message properties ourselves
+		super(data);
```
apaleslimghost, line 88: **suggestion**: rather than cloning and overriding the data object, you could construct a new object to pass to `super` (since you're setting every property anyway)? would probably be cleaner

apaleslimghost (commented):
this looks great, the rationale for rolling our own classes for this makes perfect sense to me. i have a few comments, some stylistic, some philosophical

packages/errors/lib/operational-error.js
rowanmanning: Adding these types _are_ possible in JSDoc with the following:

```js
/**
 * @typedef {Object<String, String>} OperationalErrorData
 */
```

Or in the class with

```
/**
 * @param {(String|OperationalErrorData & Object<String, String>)} [data = {}]
 */
```

However that results in VS Code's autocompletion saying that the `data` parameter is just `any` because it infers that an object which can have any string key is basically the same. So I'm torn between a ts-ignore but proper type-hinting for the `message` and `code` properties, or no type hinting but technically correct typing.

Maybe correct typing is the right way to go, as we're documenting everything as well?

packages/errors/lib/http-error.js
rowanmanning: Ah! Yes I could, it's obvious now I see it 😄  thanks

packages/errors/lib/operational-error.js
apaleslimghost: this is the type that shows up for me if i try your second suggestion, which looks correct:

Screenshot 2022-05-06 at 13 26 12

packages/errors/lib/operational-error.js
```diff
@@ -46,8 +55,23 @@ class OperationalError extends Error {
 		if (typeof data.code === 'string') {
 			this.code = OperationalError.normalizeErrorCode(data.code);
 		}
+
+		for (const [key, value] of Object.entries(data)) {
```
JSRedondo, line 59: nitpick: I _feel_ too procedural this piece of code.
Is this wrong at some point?
`this.data = Object.assign(data, ...(this.constructor.reservedKeys.map(key=>this[key])));`

jsconfig.json
```diff
@@ -8,7 +8,8 @@
 	},
 	"exclude": [
 		"coverage",
-		"node_modules"
+		"node_modules",
+		"**/*.spec.js"
```
JSRedondo, line 12: Great!

packages/errors/lib/http-error.js
rowanmanning: Ooh but actually if we define a new object, it needs to have _all_ of the properties from `data` on it anyway because `OperationalError` collects those and adds them to the `data` property. So we end up with something like this which is essentially just renaming what we already have. Maybe that's still better?

```js
// Define a new data object so that we don't modify the original
const sanitizedData = { ...data };

// Default the status code
sanitizedData.statusCode =
	typeof data.statusCode === 'number'
		? HttpError.normalizeErrorStatusCode(data.statusCode)
		: 500;

// Default the error code
sanitizedData.code =
	typeof data.code === 'string'
		? data.code
		: `HTTP_${sanitizedData.statusCode}`;

// Default the error message
sanitizedData.message =
	typeof data.message === 'string'
		? data.message
		: HttpError.getMessageForStatusCode(sanitizedData.statusCode);

// Message and code are set by the parent class, but we need
// to set the status code and message properties ourselves
super(sanitizedData);
this.statusCode = sanitizedData.statusCode;
this.statusMessage = HttpError.getMessageForStatusCode(
	sanitizedData.statusCode
);
```

Open to other suggestions!

packages/errors/README.md
```diff
@@ -58,6 +65,43 @@ OperationalError.isErrorMarkedAsOperational(new OperationalError('example messag
 OperationalError.isErrorMarkedAsOperational(new Error('example message')); // false
 ```
 
+### `HttpError`
+
+The `HttpError` class extends `OperationalError` and represents an HTTP error status. It can work in the same way as a normal error, expecting a message. In this case it will represent an HTTP `500`:
```
JSRedondo, line 84: Great decision!

packages/errors/README.md
rowanmanning: I appreciate the suggestion but yeah think I don't want to go wholesale TypeScript on this. Feels like a big commitment to a build step and change of language for the purpose of including a third-party lib. Also, in this scenario, we'd have to reimplement the things currently on the `OperationalError` class right, Like the `data`, `isOperational`, and `code` properties? Or possibly I'm misunderstanding interfaces in TS?

packages/errors/lib/http-error.js
apaleslimghost: oh, yeah, sure, that's no better really is it

JSRedondo (approved):

packages/errors/README.md
apaleslimghost: well, it would be a separate interface that both `HttpError` and `OperationalError` implement. just a way of structuring that for some kind of typed inheritance. irrelevant tho

packages/errors/lib/operational-error.js
apaleslimghost: actually yeah this is working perfectly for me?

Screenshot 2022-05-06 at 14 09 29

apaleslimghost: actually yeah this is working perfectly for me:

Screenshot 2022-05-06 at 14 10 22

apaleslimghost: ah, nice! wasn't sure to what extent typescript syntax would work here. `{[key: string]: any}` is i believe equivalent to `Record<string, any>` which imo looks nicer if it works?

rowanmanning: Nice! This also works. Thanks for all your help :)

packages/errors/lib/operational-error.js
rowanmanning: I did look at using `Object.assign` however if you look at the condition in this `for` loop you'll see it's negated, which makes the alternative a bit more complex. I can't find a simpler way to achieve the same resulting data with `Object.assign` or array iteration methods, happy if you can find something!

I've got this:

```js
this.data = Object.fromEntries(
	Object.entries(data).filter(
		// @ts-ignore TypeScript does not properly infer the constructor
		([key]) => !this.constructor.reservedKeys.includes(key)
	)
);
```

Which (once Prettier enforces code style) is the same number of lines as the `for of` except I think it's far less easy to understand at first glance.

packages/errors/README.md
rowanmanning: Too political for documentation? 😁

packages/errors/README.md
rowanmanning: I agree, I think I'd like to get to this once we have a few more modules in place and a real-world implementation. I'd lover for us to produce a go-to document for good practices in error handling. I created an issue (#15) to remind us to do this and will pop something in Jira when we're closer to being ready to write this!

alexmuller (approved):
Nice, this is looking so cool. I'm going to try running this inside an app but I don't want to block merging it for any longer so go ahead if you're happy!

PR #17, opened 2022-05-10
Build serialize-error package

rowanmanning:
This adds a serialize-error utility package. This new package exports
a utility function to serialize an error object in a way that's friendly
to loggers, view engines, and converting to JSON.

I'm writing this because we need a reliable way to serialize error data
if we tackle any of the following:

  * An error logging middleware so that thrown errors are logged
    consistently to Splunk (planned)

  * Modifications to `n-logger` to enforce more consistent logging of
    errors (not planned but likely we will need it)

  * A debugging page middleware which renders an HTML error page to help
    engineers debug while working locally (planned)

I think it makes sense to publish this as a separate package which is a
dependency of these others.

See the README in this PR for usage documentation.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,3 +1,4 @@
 {
-  "packages/errors": "0.1.1"
+  "packages/errors": "0.1.1",
+  "packages/serialize-error": "0.0.0"
 }
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ All of the packages in this monorepo are viewable in the [`packages` folder](./p
   * **[@dotcom-reliability-kit/errors](./packages/errors/#readme):**<br/>
     A suite of error classes which help you throw the most appropriate error in any situation
 
+  * **[@dotcom-reliability-kit/serialize-error](./packages/serialize-error/#readme):**<br/>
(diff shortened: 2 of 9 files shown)


rowanmanning:
Oops, ignore all the TypeScript errors – they will go away once [this commit](https://github.com/Financial-Times/dotcom-reliability-kit/pull/14/commits/d3a8b7892faa2b7fab161dc39070a72fc1ebf7a7) is merged as part of #14. An example of why it's painful to do type checking on test files 😬

alexmuller (approved):
This is working really nicely for me 👍 

I think there is a bit of potential confusion here with the interaction between `code`, `statusCode` and `status`. But I don't think we're exposing that confusion yet.

The one property that feels a little weird to me that we might need to document more in the future is `statusCode`. I _guess_ every error can have a status code and 500 is a fine default, but I suspect not all of our errors will fit neatly into HTTP statuses.

One question: are the `package-lock.json` changes required? Fine if so but I don't quite see how they relate to the PR.

alexmuller (approved):
Makes sense, thanks! Yeah worker processes are a great example of errors that don't need or want status codes.

PR #18, opened 2022-05-11
docs: link to Joyent's Error Handling documentation

alexmuller:
@rowanmanning linked to this in #ft-next-dev. It's a good intro for people who are starting to think more seriously about error handling (as we are).

I don't think this will conflict with #14 but if it will we can merge it afterwards.


rowanmanning (approved):
Good idea 👍

PR #21, opened 2022-05-11
chore: add JSDoc linting

alexmuller:
I don't feel strongly about which rules we have enabled here. From a quick look the recommended ones seem sensible and `check-access` is an obvious one that we should comply with.

`require-description-complete-sentence` seems ok although maybe a bit over-zealous?


rowanmanning (approved):
Great 🎉

PR #23, opened 2022-05-11
chore: set the ESLint JSDoc mode to TypeScript

rowanmanning:
This ensures that we can use things like type unions in our code.
See https://github.com/gajus/eslint-plugin-jsdoc#mode for documentation.


alexmuller (approved):
👍

PR #24, opened 2022-05-12
chore: don't check types in test files

rowanmanning:
I found that I was really fighting TypeScript in the tests, because it's
often useful to mock with dummy data rather than fully mocking the
expected return values of methods.

This stops us from having to litter our tests with ts-ignore comments.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -8,7 +8,8 @@
 	},
 	"exclude": [
 		"coverage",
-		"node_modules"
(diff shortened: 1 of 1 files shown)


alexmuller (approved):
My understanding is that this file is just used by local IDEs 👍

rowanmanning:
Yep we also use it to check types in the `lint` CircleCI job 👍

PR #27, opened 2022-05-16
feat: add error logging middleware

rowanmanning:
This adds error logging middleware based on the discussion in the repo
here: https://github.com/Financial-Times/dotcom-reliability-kit/discussions/26. It's relatively small because it leans on the [serialize-error](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-error#readme) and
[serialize-request](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-request#readme) packages.

You use it like this:

```js
import createErrorLogger from '@dotcom-reliability-kit/middleware-log-errors';
import express from '@financial-times/n-express';

// All your app setup and routes

app.use(createErrorLogger());
```

And all errors which get thrown in routes or passed onto Express
will end up logging all this information:

```js
{
    event: 'HANDLED_ERROR',

    // This is just the output of `@dotcom-reliability-kit/serialize-error`
    error: {
        name: 'Error',
        code: 'EXAMPLE_CODE',
        message: 'This is an example error message',
        isOperational: true,
        stack: '...',
        statusCode: 503,
        relatedToSystems: [ 'next-example' ],
        data: {}
    },

    // This is just the output of `@dotcom-reliability-kit/serialize-request`
    request: {
        id: 'xxx-xxxx-xxx',
        method: 'GET',
        url: '/categories/example?page=137',
        headers: {
            accept: '*/*',
            'content-type': null
        },
        route: {
            path: '/categories/:categoryId',
            params: {
                categoryId: 'example'
            }
        }
    },

    // Some of the easier-to-retrieve app details
    app: {
        name: 'example', // the app name as set in the `FT-App-Name` header
        region: 'EU' // the contents of the `REGION` environment variable
    }

}
```

The linked discussion contains most of the rationale behind the logging
structure. This is a first pass which we think should work well for most of
our HTTP-based apps.

### DIFF
--- a/packages/middleware-log-errors/README.md
+++ b/packages/middleware-log-errors/README.md
@@ -1,0 +1,97 @@
+
+## @dotcom-reliability-kit/middleware-log-errors
+
+Express middleware to consistently log errors. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
+
+  * [Usage](#usage)
+    * [`createErrorLogger`](#createerrorlogger)
+    * [configuration options](#configuration-options)
+      * [`includeHeaders`](#optionsincludeheaders)
+  * [Contributing](#contributing)
+  * [License](#license)
+
+
+## Usage
+
+Install `@dotcom-reliability-kit/middleware-log-errors` as a dependency:
+
+```bash
+npm install --save @dotcom-reliability-kit/middleware-log-errors
+```
+
+Include in your code:
+
+```js
+import createErrorLogger from '@dotcom-reliability-kit/middleware-log-errors';
+// or
+const createErrorLogger = require('@dotcom-reliability-kit/middleware-log-errors');
+```
+
+### `createErrorLogger`
+
+The `createErrorLogger` function can be used to generate Express middleware which logs errors to the console and Splunk via [n-logger](https://github.com/Financial-Times/n-logger). It must be added to your Express app _after_ all your application routes:
+
+```js
+const app = express();
+// App routes go here
+app.use(createErrorLogger());
+```
+
+This will automatically [serialize error objects](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-error#readme) and log them along with [a serialized HTTP request](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-request#readme) which lead to the error being thrown. The information logged looks like this:
+
+```js
+{
+    event: 'HANDLED_ERROR',
+
+    error: {
+        // See `@dotcom-reliability-kit/serialize-error` (linked above)
+        // for information about the logged properties
+    },
+
+    request: {
+        // See `dotcom-reliability-kit/serialize-request` (linked above)
+        // for information about the logged properties
+    },
+
+    app: {
+        name: 'example-app',
+        region: 'EU'
+    }
+}
+```
+
+### Configuration options
+
+Config options can be passed into the `createErrorLogger` function as an object with any of the keys below.
+
+```js
+app.use(createErrorLogger({
+    // Config options go here
+}));
+```
+
+#### `options.includeHeaders`
+
+An array of request headers to include in the serialized request object. This must be an `Array` of `String`s, with each string being a header name. It's important that you do not include headers which include personally-identifiable-information, API keys, or other privileged information. This defaults to `['accept', 'content-type']`. This option gets passed directly into [`dotcom-reliability-kit/serialize-request`](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-request#readme) which has further documentation.
+
+```js
+serializeRequest(request, {
+    includeHeaders: [
+        'accept',
+        'content-length',
+        'content-type',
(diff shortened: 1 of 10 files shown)


packages/middleware-log-errors/README.md
```diff
@@ -0,0 +1,97 @@
...
+
+```js
+import createErrorLogger from '@dotcom-reliability-kit/middleware-log-errors';
+// or
+const createErrorLogger = require('@dotcom-reliability-kit/middleware-log-errors');
+```
+
+### `createErrorLogger`
+
+The `createErrorLogger` function can be used to generate Express middleware which logs errors to the console and Splunk via [n-logger](https://github.com/Financial-Times/n-logger). It must be added to your Express app _after_ all your application routes:
```
kavanagh, line 32: Should n-logger be a peer dependency of this package?

packages/middleware-log-errors/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "license": "MIT",
+  "engines": {
+    "node": "14.x || 16.x",
+    "npm": "7.x || 8.x"
+  },
+  "main": "lib",
+  "dependencies": {
+    "@dotcom-reliability-kit/serialize-error": "^0.1.1",
+    "@dotcom-reliability-kit/serialize-request": "^0.1.0",
+    "@financial-times/n-logger": "^10.1.0"
```
kavanagh, line 21: This could feasibly be a wider range as we're only using the error method and it's signature has barely changed.

This would help to avoid the application having a different version of logger for application code.

packages/middleware-log-errors/README.md
rowanmanning: _maybe_ but unsure, I know nearly all our apps install n-logger anyway but I like the idea that you get everything you need when you install it. Gonna look at how Page Kit and Tool Kit do this.

packages/middleware-log-errors/package.json
rowanmanning: Great spot 👍  I'll have a look at what versions our apps use and widen the range to make sure nobody will _have_ to update n-logger to use this package

rowanmanning: Fixed in a new commit which switches to a range of `>=6.0.0 <11.0.0`

> I had a look across all n-logger references in `package.json` files
across the FT. It seems like nothing in Customer Products is using a
version lower than `6.x`. The only breaking changes between this
version and `10.x` are dropping support for Node.js versions - the API
doesn't change at all.

packages/middleware-log-errors/README.md
rowanmanning: I think _maybe_ peer dependencies aren't right for this case. Happy to be disagreed with but basing this on the following:

1. Page Kit and Tool Kit, from what I can see, define peer dependencies for large third-party tools like React or Webkit rather than internal dependencies (or dependencies managed by Customer Products)

2. I don't think this quite aligns with what peer dependencies are for based on the npm documentation:

    > In some cases, you want to express the compatibility of your package with a host tool or library, while not necessarily doing a require of this host. This is usually referred to as a plugin. Notably, your module may be exposing a specific interface, expected and specified by the host documentation.
    
    This middleware isn't an extension or plugin for n-logger, and so although it's unlikely to be installed in an app which doesn't already have n-logger, I think it's sensible to list it as a regular dependency.

packages/middleware-log-errors/lib/index.js
```diff
@@ -0,0 +1,61 @@
...
+			);
+		}
+	}
+
+	return (error, request, response, next) => {
+		// We add a paranoid try/catch here because it'd be really embarassing
+		// if the error logging middleware threw an unhandled error, wouldn't it
+		try {
+			const app = {
+				name: response.getHeader('ft-app-name') || null,
```
alexmuller, line 44: Would it be more reliable to read this from `process.env.SYSTEM_CODE` which Cloud Enablement requires and Tool Kit sets?

This response header _should_ get set by n-express but I don't know every case in which it might not exist. Whereas I'm pretty sure that the environment variable will be there 99.99% of the time.

alexmuller (approved):
This looks great! Going to approve with one small suggestion - feel free to ignore.

Two more general comments:

- I agree with Luke's questions on the discussion about the difference between `event` and `error.code`. I think as they're both uppercase underscore-delimited people will get confused between their purpose. However this is fine as `event` is something we set and can't be overridden.
- The comment in the README about adding this middleware "after all your application routes" could get confusing if we have other middlewares which also run late in the app. We might have to revisit that instruction.

packages/middleware-log-errors/lib/index.js
rowanmanning: Oh that's a good idea 👍  how about something like this so we have a high likelihood of getting _something_?

```js
process.env.SYSTEM_CODE || response.getHeader('ft-app-name') || null,
```

Do you think I should do it in this PR or a separate smaller feature PR?

rowanmanning: Let's go new PR, then it's less of an issue if it's not merged before I go because the current way _should_ at least work for debug-user

PR #28, opened 2022-05-17
feat: add property relatesToSystems to errors

alexmuller:
An error can be related to one or more systems. Apps that throw errors can tag their errors as being related to another system. This can include (but is not limited) when a dependency is unavailable.

I'm not sure how defensive to be about this. If a bare string is supplied I decided to convert to a one-element array so that our log searching/parsing is consistent. But we could go further - we could even use [the system code system](https://system-codes.in.ft.com/v1/check?systemCode=dotcom-debug-user) to verify if a system is valid?


alexmuller:
To be clear I think blocking an error on an HTTP request is probably a fairly bad idea, but happy to hear other opinions :)

rowanmanning (changes requested):
Code-wise I have zero changes to suggest, this looks excellent and very thorough! I'm happy with the level of defensiveness.

I think I'd consider validating system codes to be a later improvement (if we ever decide to do it) because as you hint at I think it's a potential nightmare if we introduce things to error creation (e.g. HTTP requests) that could themselves create new errors 😄 

My only changes are adding documentation in the READMEs for the two packages:

  * [Here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/packages/serialize-error/README.md#serializeerror) and [here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/packages/serialize-error/README.md#serializederror-type) in serialize-error
  * [Here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/packages/errors/README.md#operationalerror) in errors

rowanmanning (approved):
Looking great!

PR #29, opened 2022-05-17
Add a serialize-request package

rowanmanning:
This is the equivalent of the `serializeError` function but built to handle `Request` objects (either Express or plain Node.js `IncomingMessage`).

Most of the information needed to review this should be in the README (part of this PR). I'm matching our serialized request object to the current format outlined in [this discussion](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/26) so it looks like this:

```js
{
    id: 'request137',
    method: 'GET',
    url: '/fruit/feijoa',
    headers: {
        accept: '*/*'
    },
    route: {
        path: '/fruit/:fruitId',
        params: { fruitId: 'feijoa' }
    }
}
```

I decided against trying to tackle an [API Gateway request](https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html#apigateway-example-event) for now. I think it should be possible to add support for this later without requiring a breaking change.

Once this is merged, I'll refactor #27 to use this function instead of handling the serialisation itself.

### DIFF
--- a/packages/serialize-request/test/lib/index.spec.js
+++ b/packages/serialize-request/test/lib/index.spec.js
@@ -11,13 +11,15 @@ describe('@dotcom-reliability-kit/serialize-request', () => {
 				headers: {
 					accept: '*/*',
 					'content-type': 'application/json',
-					'mock-header': 'mock-param-value'
+					'mock-header': 'mock-param-value',
+					'x-request-id': 'mock-id'
 				}
 			};
 		});
 
 		it('returns the expected serialized request properties', () => {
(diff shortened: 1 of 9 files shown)


packages/serialize-request/test/lib/index.spec.js
```diff
@@ -0,0 +1,249 @@
+const serializeRequest = require('../../lib/index');
```
JSRedondo, line 1: Great tests!

JSRedondo (approved):

PR #31, opened 2022-05-25
docs: write error throwing and handling guides

rowanmanning:


JSRedondo (dismissed):

docs/getting-started/setup.md
```diff
@@ -0,0 +1,13 @@
+
+# Getting started: setup
+
+To use Reliability Kit you'll need a few things set up in advance.
+
+  1. A Node.js application (ideally built with [Express](https://expressjs.com/), but you should get some benefits even if you're not using it)
+
+  2. For your application to be running on Node.js 14 or above
+
```
JSRedondo, line 9: I would add the installation command and the quick description of the 2 packages

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+    * [Error checking](#error-checking)
+    * [Multiple try/catch blocks](#multiple-trycatch-blocks)
+  * [Handling recoverable errors](#handling-recoverable-errors)
+
+
+## Bubbling up in Express
+
+In Express, error handlers can be bound at the Application and Route level. This allows you to write lots of error handling code in one place, and you can rely on your Express error handler to do things such as [log the appropriate error detail](./logging-errors.md), forward them on to a service like Splunk, or render a custom error page.
+
+The following is not a great pattern, as Express has no chance to log the error to a central place, we're missing a lot of information that we should be capturing:
```
apaleslimghost, line 17: **nitpick**: comma splice, "we're missing..." etc should be a new sentence

```suggestion
The following is not a great pattern, as Express has no chance to log the error to a central place. We're missing a lot of information that we should be capturing:
```

apaleslimghost: now i think about it more, maybe the sentence should be restructured, because "We're missing a lot of information that we should be capturing" shouldn't be the thing that's introducting the example via the colon

ivomurrell: desperate for a ☕ here

apaleslimghost: your face is desparate for a coffee

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+    }
+});
+```
+
+Reliability Kit provides some Express error handlers which can be used to [log errors in a consistent and centralised way](./logging-errors.md).
+
+
+## Don't fear the try/catch block
+
+One of the tools we have to make sure [thrown errors should be specific](./throwing-errors.md#being-specific) is the `try`/`catch` block. In previous examples we've wrapped all our code in a single try/catch block and sent the error on to the Express error handler. If we want to be _really_ specific about what went wrong in our code, then this is often not enough.
```
apaleslimghost, line 62: **nitpick**: consistency in code quotes

```suggestion
One of the tools we have to make sure [thrown errors should be specific](./throwing-errors.md#being-specific) is the `try`/`catch` block. In previous examples we've wrapped all our code in a single `try`/`catch` block and sent the error on to the Express error handler. If we want to be _really_ specific about what went wrong in our code, then this is often not enough.
```

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+
+```js
+// E.g. GET https://your-app/fruit/feijoa
+app.get('/fruit/:name', async (request, response, next) => {
+    try {
+        // [The code in the example above, excluded for brevity]
+    } catch (error) {
+        
+        // If the name parameter is not set, let's throw a more
+        // friendly 404 error. We know that we're dealing with
```
apaleslimghost, line 106: **nitpick**: should say `400 error`

```suggestion
        // friendly 400 error. We know that we're dealing with
```

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+        // We default to passing on the error as-is, because
+        // this is probably a programmer error rather than
+        // operational and we want to let the app decide how
+        // to handle it later
+        next(error);
+    }
+});
+```
+
+With this checking in place you can be confident that your error logging (and any dashboards powered by it) will display easy to understand errors and make debugging much easier. However this checking is quite simplistic. What if we make multiple API requests, any of which can fail?
```
apaleslimghost, line 151: **nitpick**: structure

```suggestion
With this checking in place you can be confident that your error logging (and any dashboards powered by it) will display easy to understand errors and make debugging much easier. However, this checking is quite simplistic; what if we make multiple API requests, any of which can fail?
```

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+  * [What does a good error look like?](#what-does-a-good-error-look-like)
+    * [Making errors human readable](#making-errors-human-readable)
+    * [Making errors machine readable](#making-errors-machine-readable)
+    * [Adding more data](#adding-more-data)
+    * [Being specific](#being-specific)
+
+
+## Types of error
+
+Errors can be sorted into two broad categories, it's important to understand the difference because they should be handled differently by your application. These take inspiration from the [Joyent error handling documentation](https://www.joyent.com/node-js/production/design/errors), which is also worth a read.
```
apaleslimghost, line 20: **nitpick**: comma splice (also added another comma)

```suggestion
Errors can be sorted into two broad categories. It's important to understand the difference, because they should be handled differently by your application. These take inspiration from the [Joyent error handling documentation](https://www.joyent.com/node-js/production/design/errors), which is also worth a read.
```

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Operational errors
+
+An error can be considered "Operational" when they are _expected_ to occur as part of day-to-day running of an application. They normally represent that something went wrong outside of the control of the application code, e.g. a remote service failed. You can also consider an error operational when for things like end-user mistakes, e.g. filling out a form incorrectly.
```
apaleslimghost, line 24: **nitpick**: superfluous extra redundant word

```suggestion
An error can be considered "Operational" when they are _expected_ to occur as part of day-to-day running of an application. They normally represent that something went wrong outside of the control of the application code, e.g. a remote service failed. You can also consider an error operational for things like end-user mistakes, e.g. filling out a form incorrectly.
```

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+The `OperationalError` class (explained fully in the [package documentation](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#operationalerror)) allows you to throw errors which are explicitly marked as "known". Any errors created with this constructor will include an `isOperational` property set to `true`, which can be used by our tooling later to help us understand the health of a system.
+
+```js
+throw new OperationalError({
+    code: 'EXAMPLE_ERROR',
+    message: 'This is an operational error'
+});
+```
+
+This (and further error classes that which extend it) should ideally be used _anywhere_ where you understand how to handle an error. This means in your app you might do something like this:
```
apaleslimghost, line 48: **nitpick**: "anywhere where" is clunky

```suggestion
This (and further error classes that which extend it) should ideally be used _anywhere_ that you understand how to handle an error. This means in your app you might do something like this:
```

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+```js
+function uppercaseAll(array) {
+    if (!Array.isArray(array)) {
+        throw new OperationalError({
+            message: 'uppercaseAll function expects an array'
+        });
+    }
+    return array.map(item => item.toUpperCase());
+}
+```
```
apaleslimghost, line 99: **suggestion**: here, and in a couple of other places, you've introduced an example of what not to do but only explained not to do it _after_ the example. this might be confusing and cause backtracking? would be clearer to explain the examples are bad before the examples.

apaleslimghost (approved):
**praise**: this is really _really_ good documentation, and i'm glad we've got a team that cares about getting this right

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+
+  * Yourself in 6 month times when this error starts being thrown at 4am
+
+  * Your logging dashboard, which groups errors so you can easily see when specific ones spike
+
+  * Your first-line support team, who want to know why a system is crashing
+
+### Making errors human readable
+
+All errors in JavaScript support a `message` property, which is a medium-length description of the error that has occurred. `message` is ideal for catering for the "human readable" requirement of an error object, and should be written in a way that humans can read.
```
kavanagh, line 129: Should there be guidance for coming up with good code names? And perhaps a list of common names? Node.js begins all its errors with `ERR_` which helps me know the source. 

Prefix's and conventions help understand errors when seen in logs so you don't need to always refer to the source code.

Also, should they be unique?

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+// You can add any extra data you want here
+throw new OperationalError({
+    message: `A "${fruitName}" is not a valid fruit`,
+    code: 'INVALID_FRUIT',
+    input: fruitName
+});
+```
+
+### Being specific
+
```
kavanagh, line 205: When should I subclass `OperationalError` and create a system of more specific errors in my application?

kavanagh (commented):
A few quick thoughts

docs/getting-started/handling-errors.md
rowanmanning: @apaleslimghost how about this?

```suggestion
The following is not a great pattern, as Express has no chance to log the error to a central place. This means our logs could be missing information that's essential to debugging this error:
```

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+This (and further error classes that which extend it) should ideally be used _anywhere_ where you understand how to handle an error. This means in your app you might do something like this:
+
+```js
+// E.g. GET https://your-app/pokemon/oddish
+app.get('/pokemon/:name', async (request, response, next) => {
+    try {
+        // We make a request to an API, we understand that sometimes this API will fail
+        const response = await fetch(
```
sakibshaik, line 55: **suggestion:** as a good practice and from a standardisation POV, all the downstream API calls should be moved to a [helper function](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/lib/helpers.js#L3) so that the [error scenarios](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/lib/helpers.js#L29) and [validation errors](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/lib/helpers.js#L10) should be propagated back to the [calling function](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/controllers/displayPage.js#L14). the benefits of doing this will be abstracting the error handling scenarios to a common helper function and handling the errors back in the [catch section](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/controllers/displayPage.js#L97) (avoiding multiple try-catch block for multiple API calls).

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+                message: `The Pokemon API responded with a ${response.status} status`,
+                relatesToSystems: ['pokeapi']
+            });
+        }
+
+        response.send(await response.json());
+
+    } catch (error) {
+        // Pass any errors onto the Express error handler
+        next(error);
```
sakibshaik, line 73: **suggestion:** also, it would be clean to have a generic [error handler](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/controllers/displayPage.js#L97) for [operational errors](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/controllers/errorHandler.js#L5)

sakibshaik (commented):
thanks for putting this together, its a good starting point for error handling

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+
+```js
+// This is better because Express will get hold of the error, potentially
+// log it, and render a page with a `500` status code
+app.get('/', (request, response) => {
+    exampleFunctionWhichThrowsAnError();
+});
+```
+
+For `async` route handlers, it's very important that you either use a module like [async-express-errors](https://www.npmjs.com/package/express-async-errors) (to avoid app crashes) or wrap your handler in an outer try/catch block:
```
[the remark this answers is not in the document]
rowanmanning: I see where you're coming from but I think because we can't _mandate_ that teams user `express-async-errors` I'd like to keep examples assuming that they don't. I think it might be worth adding an entirely new documentation page outside of the "getting started" guide which covers `async` route handlers and then linking to it from here.

I think I can add in a few clarifying messages to explain to the reader that we're making the assumption that they're not monkey-patching Express, and also that if they _are_ then they can ignore the `next` pattern.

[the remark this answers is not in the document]
rowanmanning: I've updated the docs to better explain this 👍  thanks. See the most recent commit! Feel free to unresolve if you think we need more

docs/getting-started/throwing-errors.md
rowanmanning: I think this is covered by the [bubbling up in express](https://github.com/Financial-Times/dotcom-reliability-kit/blob/getting-started-guide/docs/getting-started/handling-errors.md#bubbling-up-in-express) section. It's better to pass the errors onto `next` because they will be caught by an Express error handler later.

I haven't documented it yet because it's in-progress, but [we're working on some built-in error handlers](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/26) which _do_ properly handle errors, so you'll be able to add the following to your app:

```js
const createErrorLoggingMiddleware = require('@dotcom-reliability-kit/middleware-log-errors');

app.use(createErrorLoggingMiddleware());
```

In order to take advantage of these we'd have to be using the `next` function in a route rather than something custom because otherwise the errors won't make it to central error handlers and will be output by each route.

The example you give still relies on each route importing an error handler function when it could be handing over to Express instead and avoiding having to remember to pass on all the details like `UUID` or `response`.

docs/getting-started/throwing-errors.md
rowanmanning: Oh also I couldn't help but notice in [this example](https://github.com/Financial-Times/next-newsletters-page/blob/abd26fbb5233b75b1b4f56b76c5adf1f336d7666/server/lib/helpers.js#L10) that your app is throwing a plain object with a `message` property rather than an `Error`. This means that any errors which get logged will not be able to include an error `stack` which is really helpful in debugging.

I'm going to add a section to the documentation which explains this, but in the meantime the two ESLint rules related to throwing and rejecting non-`Error` objects are useful:

  * https://eslint.org/docs/rules/no-throw-literal
  * https://eslint.org/docs/rules/prefer-promise-reject-errors

sakibshaik: > I don't want to dictate _everything_ that we do with error handling

I wouldn't say it is dictating, you can have it like: here are the two ways of doing it and chose your way. Obviously, there are tradeoffs between both approaches. so the consumer can decide on the approach

docs/getting-started/throwing-errors.md
sakibshaik: middleware-log-errors npm approach makes sense. in that case, I would resolve this conversation.

docs/getting-started/throwing-errors.md
rowanmanning: I've now documented this in the [throwing errors](https://github.com/Financial-Times/dotcom-reliability-kit/blob/getting-started-guide/docs/getting-started/throwing-errors.md#using-error-objects) guide, thanks for the suggestion!

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,289 @@
...
+    * [Adding more data](#adding-more-data)
+    * [Being specific](#being-specific)
+
+
+## Types of error
+
+Errors can be sorted into two broad categories, it's important to understand the difference because they should be handled differently by your application. These take inspiration from the [Joyent error handling documentation](https://www.joyent.com/node-js/production/design/errors), which is also worth a read.
+
+### Operational errors
+
```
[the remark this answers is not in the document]
rowanmanning: I 100% agree that these should have tests 😄  but I feel like we have to draw the line on what we document somewhere. It feels like a job for more central place (the wiki?) to outline our testing practices. I'm painfully aware we're already above 2,000 words in this document, not including code examples!

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+        // pass the error onto the next route handler so that
+        // it can be logged and an appropriate error page can
+        // be rendered
+        next(error);
+    }
+});
+```
+
+Reliability Kit provides some Express error handlers which can be used to [log errors in a consistent and centralised way](./logging-errors.md).
+
```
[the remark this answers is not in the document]
rowanmanning: I do think we should document this stuff for sure, but maybe not in the getting started guide? The more I think about the more I want to add, maybe we need a separate FAQ-like document for specific cases? Or maybe it's OK for teams to make their own call for now and we can document the patterns that come out of this?

docs/getting-started/handling-errors.md
```diff
@@ -0,0 +1,326 @@
...
+    }
+});
+```
+
+With this checking and multiple `try`/`catch` blocks in place you can be confident about exactly which part of your app is failing. If an API is down then you get a message which specifically outlines which. Your tooling/reporting can also confidently say X% of requests fail because this specific API is down a lot and you can prioritise some work to address this.
+
+
+## Handling recoverable errors
+
+Sometimes an error in your code is recoverable, and a page can continue to be rendered even if something failed. We call these recoverable errors. Even when an error is recoverable, it's still useful to have logs to help us understand how frequently certain parts of a website are missing.
```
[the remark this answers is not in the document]
rowanmanning: Yeah definitely, once I'm done with error logging I want to make sure we provide _something_ more concrete for logging "grey failure"/recoverable errors, this doc will be edited further once I PR new packages in to make sure it stays up to date with what we have codified.

docs/getting-started/throwing-errors.md
rowanmanning: I've put in a small note on error code conventions. I think the answer is that I don't know whether we should define conventions yet. I think it might be better if solid conventions come out of the teams who are implementing errors. It's fine for codes to change over time as part of a review that happens later 🙂

docs/getting-started/throwing-errors.md
rowanmanning: Yep I don't know the answer to this 😂  I think again let's wait and see what errors the teams find that they need. It may be that if the same error types crop up consistently then they could graduate to the `@dotcom-reliability-kit/errors` package?

alexmuller (approved):
These docs are great - very readable and easy to follow along with. One small typo but I can fix separately if you want to get this merged.

docs/getting-started/throwing-errors.md
```diff
@@ -0,0 +1,316 @@
...
+# Getting started: throwing errors
+
+Throwing good errors is key to producing a reliabile application. Most of Reliability Kit's packages rely on sensible errors being thrown if a fault is encountered in your app, and without reviewing your existing errors you won't get as much value from using it.
```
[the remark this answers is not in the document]
rowanmanning:
```suggestion
Throwing good errors is key to producing a reliable application. Most of Reliability Kit's packages rely on sensible errors being thrown if a fault is encountered in your app, and without reviewing your existing errors you won't get as much value from using it.
```

PR #35, opened 2022-05-31
feat: use the SYSTEM_CODE environment variable

rowanmanning:
This switches the logging middleware to prefer the `SYSTEM_CODE`
environment variable over the "ft-app-name" response header. This means
that we're more reliably going to have an app name.


alexmuller (approved):
The jump in line numbers on the diff caught me way off guard (thanks git) but this looks good 👍

PR #44, opened 2022-06-13
docs: add an Express example app

rowanmanning:
This serves two purposes:

  1. We now have an example application which complements the
     documentation in the repo

  2. We can run end-to-end tests using the example application, some
     have been added as part of this work.


JSRedondo (approved):
I think this PR is great! Aren't examples the best way to start learning about how to do something? 😉

alexmuller (approved):
Looks great, nice and easy to use.

Coverage of the example app is <100% but that doesn't matter because Jest is smart enough to have `coverageThreshold` with a path option. Nice.

PR #46, opened 2022-07-01
Set up Dependabot

rowanmanning:
We need dependabot to be configured for this to work.
[Docs on the config options here](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


alexmuller (approved):
Makes sense.

- I think this will only update `devDependencies` in the root `package.json` and not the individual package dependencies?
- I'm not sure where we got to in the wider Customer Products/FT on the whole Dependabot vs Renovate situation
- We may need to update our CircleCI config so that a bunch of Dependabot PRs don't clog up CI

PR #49, opened 2022-07-05
chore: add label filter to package bug pages

rowanmanning:


alexmuller (approved):
Weird that the previous change didn't work, but yes let's try it

PR #56, opened 2022-07-05
chore: add dependabot config for packages

rowanmanning:
This means that Dependabot should now keep the individual packages up to
date as well as the central dev dependencies.


alexmuller (approved):
Nice! I'm tempted to suggest that `scripts/create-package.js` could parse the YAML and update it but I agree that one line in the README is sufficient :trollface:

PR #71, opened 2022-07-11
feat: require n-logger v10.2.0

rowanmanning:
n-logger v10.2.0 adds in the ability to switch to Heroku log drains
rather than sending logs to Splunk manually. In order to support this
feature we need to limit the version range to be v10.2.0 or above
otherwise the feature will be present inconsistently.

This is not a breaking change because the major versions of n-logger
between 6 and 10 have no impact on this package:

  - v7.0.0 of n-logger only supports Node.js 14+ (no change for us)
  - v8.0.0 of n-logger re-adds support for Node.js 12
  - v9.0.0 of n-logger deprecates the SYSTEM_CODE environment variable
    (but then undeprecates it in v9.0.1)
  - v10.0.0 of n-logger only supports Node.js 14+ (no change for us)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1136,9 +1136,9 @@
       }
     },
     "node_modules/@financial-times/n-logger": {
-      "version": "10.1.0",
-      "resolved": "https://registry.npmjs.org/@financial-times/n-logger/-/n-logger-10.1.0.tgz",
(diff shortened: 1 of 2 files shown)


alexmuller (approved):
Looks right to me!

PR #83, opened 2022-07-21
chore: release main

next-team:
:robot: I have created a release *beep* *boop*
---

<details><summary>log-error: 1.2.0</summary>

### Features

* log an error summary in the message property ([a4484d0](https://github.com/Financial-Times/dotcom-reliability-kit/commit/a4484d0aca72bf730f1ea2df2773f33e346a3f53))

### Documentation Changes

* correct a missing documentation change ([1c32b26](https://github.com/Financial-Times/dotcom-reliability-kit/commit/1c32b26f213290794de289fd4fb6754e3b46af00))
* manually fix the broken changelogs ([7249a61](https://github.com/Financial-Times/dotcom-reliability-kit/commit/7249a6177c7bed3ea4b53254aef259b9b5fe2d4c))

### Miscellaneous

* **dev-deps:** update release-please to 13.18.6 ([5078c57](https://github.com/Financial-Times/dotcom-reliability-kit/commit/5078c57d651afc624d1fac0934f983cc0d5bd428))
* tweak some internal naming for clarity ([c3c6ac6](https://github.com/Financial-Times/dotcom-reliability-kit/commit/c3c6ac6a1e00ae3870eb57a38b86d6b5c6ff3bc3))
</details>

<details><summary>middleware-log-errors: 1.1.0</summary>

### Features

* log an error summary in the message property ([a4484d0](https://github.com/Financial-Times/dotcom-reliability-kit/commit/a4484d0aca72bf730f1ea2df2773f33e346a3f53))

### Documentation Changes

* correct a missing documentation change ([1c32b26](https://github.com/Financial-Times/dotcom-reliability-kit/commit/1c32b26f213290794de289fd4fb6754e3b46af00))
* manually fix the broken changelogs ([7249a61](https://github.com/Financial-Times/dotcom-reliability-kit/commit/7249a6177c7bed3ea4b53254aef259b9b5fe2d4c))

### Miscellaneous

* **dev-deps:** update release-please to 13.18.6 ([5078c57](https://github.com/Financial-Times/dotcom-reliability-kit/commit/5078c57d651afc624d1fac0934f983cc0d5bd428))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @dotcom-reliability-kit/log-error bumped from ^1.1.0 to ^1.2.0
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #84, opened 2022-07-21
Fix release please

rowanmanning:
This is annoying, but I dug into the code for Release Please and it gets the manifest config directly from GitHub rather than the local file so our change needs to be merged into `main` before we can progress beyond Release Please v13.18.6.

This addresses the broken changelog files and also makes sure we're using a version of Release Please that works.


alexmuller (approved):
Thanks - hopefully it will automatically fix the open PR #83?

PR #89, opened 2022-07-28
Apply documentation amends

andygout:
A few typo fixes and suggestions to the documentation having read through it.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -133,7 +133,7 @@ The prefix is used to calculate the semver release level:
 | feat      | a feature has been added                               | `minor`       |
 | fix       | a bug has been patched                                 | `patch`       |
 | docs      | a change to documentation                              | none          |
-| chore     | any changes that don't impact end users of the library | none          |
+| chore     | repo maintenance and support tasks                     | none          |
 
 Indicate a breaking change by placing an `!` between the type name and the colon, e.g.
 
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -23,9 +23,9 @@ We're glad you want to contribute to Reliability Kit!
 
 In order to contribute and make changes to this repository, you'll need some software installed:
 
-  * [Node.js](https://nodejs.dev/) v14 or higher
+  * [Node.js](https://nodejs.dev) on a version defined in `package.json`: `engines.node`
 
-  * [npm](https://www.npmjs.com/) v7 or higher (usually bundled with Node.js)
+  * [npm](https://www.npmjs.com) on a version defined in `package.json`: `engines.npm` (usually bundled with Node.js)
 
 ### Optional
 
@@ -141,11 +141,6 @@ Indicate a breaking change by placing an `!` between the type name and the colon
 feat!: add a breaking feature
 ```
 
-or
-
-```
-chore!: drop support for Node.js v12
-```
 
 ### Commit linting
 
--- a/docs/getting-started/handling-errors.md
+++ b/docs/getting-started/handling-errors.md
@@ -95,7 +95,7 @@ The Express documentation covers [writing centralised error handlers](https://ex
 
 ## Don't fear the try/catch block
 
-One of the tools we have to make sure [thrown errors should be specific](./throwing-errors.md#being-specific) is the `try`/`catch` block. In previous examples we've wrapped all our code in a single `try`/`catch` block and sent the error on to the Express error handler. If we want to be _really_ specific about what went wrong in our code, then this is often not enough.
+One of the tools we have to make sure [thrown errors are specific](./throwing-errors.md#being-specific) is the `try`/`catch` block. In previous examples we've wrapped all our code in a single `try`/`catch` block and sent the error on to the Express error handler. If we want to be _really_ specific about what went wrong in our code, then this is often not enough.
 
 We have a couple of different ways that we update our `try`/`catch` to help us to handle errors and rethrow more specific ones. We can use either or both of these.
 
@@ -121,7 +121,7 @@ app.get('/fruit/:name', async (request, response, next) => {
 });
 ```
 
-In this scenario we have the following possible errors which we know can make it into our final try/catch:
+In this scenario we have the following possible errors which we know can make it into our final `try`/`catch`:
 
   * The `name` parameter is an empty string, throwing an `AssertionError`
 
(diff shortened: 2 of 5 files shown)


docs/contributing.md
```diff
@@ -23,9 +23,9 @@ We're glad you want to contribute to Reliability Kit!
 
 In order to contribute and make changes to this repository, you'll need some software installed:
 
-  * [Node.js](https://nodejs.dev/) v14 or higher
+  * [Node.js](https://nodejs.dev) on a version defined in `package.json`: `engines.node`
 
-  * [npm](https://www.npmjs.com/) v7 or higher (usually bundled with Node.js)
+  * [npm](https://www.npmjs.com) on a version defined in `package.json`: `engines.npm` (usually bundled with Node.js)
```
andygout, line 28: I always find the documentation is usually the place that forgets to get updated as part of a Node upgrade as that it is more update-proof for the docs to simply referencing somewhere else in the codebase where the Node version depends on being updated.

docs/contributing.md
```diff
@@ -141,11 +141,6 @@ Indicate a breaking change by placing an `!` between the type name and the colon
 feat!: add a breaking feature
 ```
 
-or
-
-```
-chore!: drop support for Node.js v12
```
andygout, line 147: I thought this example might cause confusion with the definition of `chore` - dropping support for Node.js v12 could hypothetically impact end users of the library, e.g. for consumers that still use Node v12.

Is there another type that covers this sort of work or should we revise the definition of `chore` accordingly?

docs/getting-started/handling-errors.md
```diff
@@ -147,7 +147,7 @@ app.get('/fruit/:name', async (request, response, next) => {
             return next(new HttpError({
                 statusCode: 400,
                 code: 'INVALID_FRUIT_NAME',
-                message: 'Please provide a fruit name'
+                message: 'A fruit name was not provided'
```
andygout, line 150: `Please provide a fruit name` felt like an instruction to the end user of a GUI, and hypothetically to the app could just take this message value and pass it on for that purpose, but it feels like the error message of an HTTP error should be more of a statement as to the cause of the error?

docs/getting-started/handling-errors.md
```diff
@@ -146,8 +146,8 @@ app.get('/fruit/:name', async (request, response, next) => {
         if (error.code === 'ERR_ASSERTION') {
             return next(new HttpError({
                 statusCode: 400,
-                code: 'INVALID_FRUIT_NAME',
-                message: 'Please provide a fruit name'
+                code: 'FRUIT_NAME_NOT_PROVIDED',
```
andygout, line 149: The error is thrown if the string is empty, while `INVALID_FRUIT_NAME` suggests that a string could have been provided but isn't a fruit (e.g. `'elephant'`), though this is only validated by the subsequent API call.

`FRUIT_NAME_NOT_PROVIDED` is more specific in detailing that a non-whitespace string but not provided whatsoever.

docs/contributing.md
alexmuller: Interesting point! I agree I can't untangle the definition of chore ("any changes that don't impact end users of the library") which could also be a breaking change. Would be interested to hear what Rowan thinks :)

docs/contributing.md
rowanmanning: Good shout, one less place to update 👍

docs/getting-started/throwing-errors.md
```diff
@@ -18,7 +18,7 @@ Throwing good errors is key to producing a reliable application. Most of Reliabi
 
 ## Types of error
 
-Errors can be sorted into two broad categories. It's important to understand the difference, because they should be handled differently by your application. These take inspiration from the [Joyent error handling documentation](https://www.joyent.com/node-js/production/design/errors), which is also worth a read.
+Errors can be sorted into two broad categories. It's important to understand the difference, because they should be handled differently by your application. These take inspiration from the [Joyent error handling documentation](https://web.archive.org/web/20220223020910/https://www.joyent.com/node-js/production/design/errors), which is also worth a read.
```
alexmuller, line 21: Sad! I wonder why they got rid of it 👀

docs/contributing.md
rowanmanning: Ah yeah I see what you mean 🤔 I think redefine `chore` because dropping Node.js version support does technically come under that heading (it's not a feature, it's not a fix, and it's not documentation).

Shall we go with `repo maintenance and support tasks`? I feel like maybe that covers it

docs/getting-started/throwing-errors.md
```diff
@@ -212,7 +212,7 @@ Your error codes should uniquely identify a _type_ of error your app can throw.
 
 ### Adding more data
 
-It's also sometimes useful to add extra details to an error so that further debugging of them can be done once we've determined what _type_ of error they are using `code`. Useful information might be the actual value of any input which caused an error, or further details about application state (e.g. the ID of the user this error occurred for).
+It's also sometimes useful to add extra details to an error so that further debugging of them can be done once we've determined what _type_ of error they are using via the `code` value. Useful information might be the actual value of any input which caused an error, or further details about application state (e.g. the ID of the user this error occurred for).
```
alexmuller, line 215: __question:__ did you mean to delete the "using" here to read "once we've determined what type of error they are via the code value"?

docs/getting-started/throwing-errors.md
rowanmanning, line 21: Nice 👏  this feels more stable

rowanmanning (approved):
This is great, thanks very much for making the improvements! Some glaring typos in there 🙈

alexmuller (commented):
This is great, thanks

docs/contributing.md
andygout: `repo maintenance and support tasks` works for me 😄

docs/getting-started/throwing-errors.md
andygout: I don't think I have removed "using"?

The edited sentence reads:
> It's also sometimes useful to add extra details to an error so that further debugging of them can be done once we've determined what _type_ of error they are using via the `code` value.

docs/contributing.md
andygout: Updated in [this commit](https://github.com/Financial-Times/dotcom-reliability-kit/commit/eab86bde7703f0245bc39c027d0f04dac52187f8).

PR #92, opened 2022-07-29
feat: add new app properties to error logs

rowanmanning:
This adds new app-related properties to error logs. We decided these
would be useful to help us debug issues and will bring logs in Splunk
closer to Sentry in terms of the information they hold.

These are the added properties:

  - `app.commit`: set to `process.env.HEROKU_SLUG_COMMIT`
  - `app.nodeVersion`: set to `process.versions.node`
  - `app.releaseDate`: set to `process.env.HEROKU_RELEASE_CREATED_AT`

Some of these rely on the Heroku Dyno Metadata labs feature but we
recommend use of this in all our apps anyway:
https://devcenter.heroku.com/articles/dyno-metadata

**Note to reviewers:** most of the changes here are to the tests, which
all need to be updated when we add new properties to fields. I think the
tests themselves need less of a review as it's a lot of repetition.

This resolves FTDCS-248.

### DIFF
--- a/packages/log-error/README.md
+++ b/packages/log-error/README.md
@@ -54,8 +54,11 @@ This will automatically [serialize error objects](https://github.com/Financial-T
     },
 
     app: {
+        commit: '137da65185397a7d699ed54c3052d10d83e82137',
         name: 'example-app',
-        region: 'EU'
+        nodeVersion: '16.16.0',
+        region: 'EU',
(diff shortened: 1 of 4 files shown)


andygout (approved):
I am guessing that `process.versions.node` is guaranteed to always be present (unlike the `process.env` values)?

PR #98, opened 2022-08-03
feat: add OperationalError cause property

andygout:
FTDCS-250

Jira card: https://financialtimes.atlassian.net/jira/software/c/projects/FTDCS/boards/1426?modal=detail&selectedIssue=FTDCS-250

> **Rowan Manning:** @andygout [pointed out](https://financialtimes.slack.com/archives/CPNKK12S1/p1659013086713049?thread_ts=1659003590.492259&cid=CPNKK12S1) in one of my examples for handling errors we lose some information about the original error. I’ve had a potential solution in mind for this for a while and it’s worth getting down on paper.

> I think that OperationalError should have a new first-class property named cause. This expects another Error object and marks that error as the root cause of the operational error. This will allow us to use generic messages in some places but still capture the original error information. Using the example from the error handling docs, we could update it to be:

```js
try {
    fruit.stockLevel = await fruitApiClient.getStockLevel(fruit.id);
} catch (error) {
    console.warn({
        event: 'RECOVERABLE_ERROR',
        error: serializeError(new OperationalError({
            code: 'FRUIT_STOCK_LEVEL_FAILED',
            message: `The Fruit API did not return a stock level for ID ${fruit.id}`,
            relatesToSystems: ['fruit-api'],
            cause: error // This is the new code
        }))
    });
    fruit.stockLevel = null;
}
```

> We’d also want to update our logger so that it exposes this information and also serialises any errors on the cause property, e.g. the log output could change to:

```js
{
    event: 'RECOVERABLE_ERROR',
    message: 'OperationalError: The Fruit API did not return a stock level for ID X',

    error: {
        // missing some properties for brevity
        name: 'OperationalError',
        code: 'FRUIT_STOCK_LEVEL_FAILED',
        message: 'The Fruit API did not return a stock level for ID X',
        
        cause: {
            name: 'Error',
            code: null,
            message: 'HTTP 503'
        }
    },

    app: {
        // etc
    }
}
```

### DIFF
--- a/docs/getting-started/handling-errors.md
+++ b/docs/getting-started/handling-errors.md
@@ -337,10 +337,10 @@ app.get('/fruit/:name', async (request, response, next) => {
             console.warn({
                 event: 'RECOVERABLE_ERROR',
                 error: serializeError(new OperationalError({
-                    statusCode: error.statusCode,
                     code: 'FRUIT_STOCK_LEVEL_FAILED',
                     message: `The Fruit API did not return a stock level for ID ${fruit.id}`,
-                    relatesToSystems: ['fruit-api']
+                    relatesToSystems: ['fruit-api'],
+                    cause: error
                 }))
             });
             fruit.stockLevel = null;
--- a/docs/getting-started/throwing-errors.md
+++ b/docs/getting-started/throwing-errors.md
@@ -228,6 +228,20 @@ throw new OperationalError({
 });
 ```
 
+You might create an `OperationalError` instance in reaction to an error that has been caught. The root cause error and all the diagnostic information it contains can be included in the `OperationalError` instance by setting it as the value of the `cause` property:
+
+```js
+try {
+    fruit.stockLevel = await fruitApiClient.getStockLevel(fruit.id);
+} catch (error) {
+    throw new OperationalError({
+        message: `A "${fruitName}" is not a valid fruit`,
+        code: 'INVALID_FRUIT',
+        cause: error
+    });
+}
+````
+
 ### Being specific
 
 When you're working on more complex code than our previous examples and catching errors in places where they could have been caused by multiple things, it's important to throw _specific_ errors which identify the cause rather than a generic one.
--- a/packages/serialize-error/lib/index.js
+++ b/packages/serialize-error/lib/index.js
@@ -12,6 +12,8 @@
  *     A human readable message which describes the error.
  * @property {boolean} isOperational
  *     Whether the error is operational, as in it's an error we expect sometimes as part of running the application.
+ * @property {(Error | null)} cause
+ *     The root cause error instance.
  * @property {(string | null)} stack
  *     The full error stack.
  * @property {(number | null)} statusCode
@@ -61,6 +63,11 @@ function serializeError(error) {
 		errorProperties.relatesToSystems = error.relatesToSystems;
 	}
 
+	// If set, error cause (which in turn is an error instance) is serialized
+	if (error.cause && error.cause instanceof Error) {
+		errorProperties.cause = serializeError(error.cause);
+	}
+
 	// Only include error stack if it's a string
 	if (typeof error.stack === 'string') {
 		errorProperties.stack = error.stack;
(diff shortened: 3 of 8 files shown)


docs/getting-started/handling-errors.md
```diff
@@ -337,10 +337,10 @@ app.get('/fruit/:name', async (request, response, next) => {
             console.warn({
                 event: 'RECOVERABLE_ERROR',
                 error: serializeError(new OperationalError({
-                    statusCode: error.statusCode,
                     code: 'FRUIT_STOCK_LEVEL_FAILED',
                     message: `The Fruit API did not return a stock level for ID ${fruit.id}`,
-                    relatesToSystems: ['fruit-api']
+                    relatesToSystems: ['fruit-api'],
+                    cause: error
```
rowanmanning, line 343: **praise:** nice 👏 this is a much better example

docs/getting-started/throwing-errors.md
```diff
@@ -228,6 +228,20 @@ throw new OperationalError({
 });
 ```
 
+You might create an `OperationalError` instance in reaction to an error that has been caught. The root cause error and all the diagnostic information it contains can be included in the `OperationalError` instance by setting it as the value of the `cause` property:
```
rowanmanning, line 231: **praise:** great explanation and example

packages/serialize-error/test/lib/index.spec.js
```diff
@@ -84,6 +85,30 @@ describe('@dotcom-reliability-kit/serialize-error', () => {
...
+				const rootCauseErrorInstance = new Error(
+					'mock root cause error message'
+				);
+				error.cause = rootCauseErrorInstance;
+				expect(serializeError(error)).toMatchObject({
+					cause: serializeError(rootCauseErrorInstance)
+				});
+				expect(serializeError(error).cause).toMatchObject({
+					message: 'mock root cause error message'
+				});
```
andygout, line 99: I could have gone to town on testing the serialisation of the `cause` error to provide the same coverage that is given for the subject (i.e. containing) error, e.g. when the `cause` value is the result of `serialiseError()` being called with:
- an error object
- en error-like object
- a string
- a number
- an array

However, the `serialiseError()` is already being tested so thoroughly by the subject error, so ensuring that the root cause error `message` value is as expected felt to be a sufficient measure for testing this.

andygout:
> We’d also want to update our logger so that it exposes this information

From what I can tell, the `cause` value of error should be surfaced via the inclusion of the `error` property (in which it is situated) and that no changes are required to `packages/log-error/lib/index.js` but please let me know if I have misunderstood something.

packages/serialize-error/lib/index.js
```diff
@@ -61,6 +63,11 @@ function serializeError(error) {
 		errorProperties.relatesToSystems = error.relatesToSystems;
 	}
 
+	// If set, error cause (which in turn is an error instance) is serialized
+	if (error.cause && error.cause instanceof Error) {
+		errorProperties.cause = serializeError(error.cause);
```
rowanmanning, line 68: **thought:** maybe down the line we need to think about whether to limit the recursion here? I don't think in this PR but I guess it's a possible vector for errors to creep in. Though tbh I think you'd have to try quite hard to make this infinitely recurse 🤷 anyway not something to address now

rowanmanning (changes requested):
This is looking great, really excellent 🙂 and I think it'll make `OperationalError` a lot more useful when it starts appearing in apps.

My one change request is that we need to update the READMEs for a few of the packages:

* In the serialize-error README [here](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-error#serializederror-type) and [here](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/serialize-error#serializeerror) (also don't forget to update the table of contents which is manual for now)
* In the errors README [here](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#operationalerrorrelatestosystems) (also the ToC)

rowanmanning (approved):
Looks really great 👍 thanks Andy! Feel free to merge

PR #102, opened 2022-08-04
feat: add a middleware-render-error-info package

rowanmanning:
This makes a start on FTDCS-184, adding a new middleware function which
renders an error information page for debugging. This is the absolute
bare-bones of the module so far and it only provides some basic error
information.

There are also some unanswered questions, mostly whether this will play
nicely with the n-raven and n-express error handlers. I think this will
need some testing once we have a basic 0.1.0 release of this package.

This does not include documentation yet, I think I'd rather work on that
once we've nailed down exactly what the behaviour is and whether we
include any config options.

The commit is tagged to release as 0.1.0, so this should be OK to merge
even though it's incomplete. That way we'll all be able to incrementally
add features rather than having a long-running PR.

---

I tested roughly with the example Express app, it currently looks like this:

Screenshot 2022-08-04 at 13 30 47

### DIFF
--- a/examples/express/test/index.spec.js
+++ b/examples/express/test/index.spec.js
@@ -6,6 +6,10 @@ jest.mock('@financial-times/n-logger', () => ({
 	}
 }));
 
+// The node enironment be set to "test" to get around the error info renderer
+// so that our JSON error handler triggers
+process.env.NODE_ENV = 'test';
+
 process.env.SYSTEM_CODE = 'reliability-kit-express-example';
 process.env.REGION = 'test';
 
--- a/examples/express/test/index.spec.js
+++ b/examples/express/test/index.spec.js
@@ -6,7 +6,7 @@ jest.mock('@financial-times/n-logger', () => ({
 	}
 }));
 
-// The node enironment be set to "test" to get around the error info renderer
+// The node environment is set to "test" to get around the error info renderer
 // so that our JSON error handler triggers
 process.env.NODE_ENV = 'test';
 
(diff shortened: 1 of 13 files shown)


alexmuller (approved):
This is a really exciting thing to start. Two thoughts, neither of which block this PR:

- I wonder if we should explain to developers why they're seeing this page so that they don't have to go digging for it
  > Reliability Kit has generated this debug page because your `NODE_ENV` is `whatever`
- Perhaps we want some kind of end-to-end test to show how this interacts with n-raven and make sure that exceptions still flow to the right places.

rowanmanning:
Agreed on both 👍 I think I'll do the n-raven test next because it'll help me to understand how they interact. A message on the debug page would be great too

examples/express/test/index.spec.js
```diff
@@ -6,6 +6,10 @@ jest.mock('@financial-times/n-logger', () => ({
 	}
 }));
 
+// The node enironment be set to "test" to get around the error info renderer
```
andygout, line 9:
```suggestion
// The node environment is set to "test" to get around the error info renderer
```

packages/middleware-render-error-info/README.md
```diff
@@ -0,0 +1,4 @@
+
+## @dotcom-reliability-kit/middleware-render-error-info
+
+Express middleware to render error information in a way that makes debugging easier. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
```
andygout, line 4:
```suggestion
Express middleware to render error information in a browser in a way that makes debugging easier. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
```

To make explicit where the error information will be viewable (although 'render' does already heavily imply this).

andygout (approved):
I have made a couple of very minor non-blocking suggestions.

PR #103, opened 2022-08-04
feat: design a cute logo

rowanmanning:
| ✨ Current ✨ | Starting Point |
| --- | --- |
|  Rock Logo | Rock Logo |

This is a rough start at creating a cute logo for Reliability Kit. It's
non-essential to the running, but we want something that:

  1. Represents "reliability"

  2. Is cute and will make a desirable sticker :eyes:

The decision making process so far has been:

  1. I asked @apaleslimghost on Slack what emoji we could use to
     represent "reliability", because it was a low-effort way to make
     the repo more interesting. She came up with 🪨 which is a nice
     choice

  2. Much later @alexmuller raised the idea of using stickers to
     encourage people to use Reliability Kit. I thought it might be a
     nice way to increase the hype

  3. The idea wouldn't leave my head, so I spent some evening time
     mocking something up

  4. After some feedback from @andygout and Emma B, I made a
    few tweaks to refine the idea

## Feedback

I'm looking for feedback at pretty much any level, including but not
limited to:

  * This is a bad idea, let's start again or not have a logo
  * General tweaks to the shape/style
  * Comments on the facial expression – I was going for
     "tough" or "determined" but not "angry", does it capture this?
    (already changed this a lot, I'm happier with where we are now)
  * Colouring choices, it's a slightly green grey, does that work?
  * Is the shading/detail rock-like enough? (already made some
    tweaks but the question still stands)
  * <del>Does it need some line-work / an outline?</del> <ins>I updated to
    have an outline, I think it works better</ins>
  * <del>Do you like it with/without the plaster?</del> <ins>I updated to remove
    the plaster, I agree it looks better without</ins>
  * Does it need a name? :3 (Andy suggests Rocky)

## Other assets

There's also a mono version which fits to a 32x32 pixel grid
that works better at smaller sizes:

Rock Sketches
Rock Sketches

There's also a version with the project name alongside for
use in the README. I decided to use [Roboto](https://en.wikipedia.org/wiki/Roboto) for this as it's
relatively close to Metric and is free to use:

Reliability

I couldn't be bothered to scan properly but here are a few
early sketches:

Rock Sketches


rowanmanning:
Some feedback I'm gonna think about and/or action from Emma B in design:

* A rock is good to represent reliability "steady as a rock" and both Andy and I immediately get "solid as a rock" in our heads when we see it
* It might seem a little angry, possibly needs to be more calm/pleasant and approachable (changing eyebrows might be a low-effort way to do this or switching the cheeks to circles rather than ovals)
* Agree with Andy that we should ditch the plaster, didn't quite make sense
* Not sure if the shading quite works, might be worth a revisit

andygout:
> Does it need a name?

Surely it's Rocky? 😆

rowanmanning:
There's now a version with the Reliability Kit name incorporated.
This is how it appears in the README:

Screenshot 2022-08-08 at 15 38 34

andygout (approved):
I'm guessing that `logos/src/rock-mono.svg` intentionally has a differently shaped mouth (presumably so that it is more clearly visible as a smaller icon)?

rowanmanning:
Yep I tried using the same shape and just couldn't get it working at smaller sizes. Happy for someone else to have a go down the line :)

alexmuller (approved):
This is great, I love it 🪨 

My understanding is that because this is a separate (unpublished) library the `postinstall` only applies to installation in this repo and not when somebody installs a Reliability Kit package 😅

alexmuller:
Do you want to edit `/README.md` to include it as part of this PR?

rowanmanning:
Yep that's right, the postinstall will only run when `npm i` is run from within the root of Reliability Kit, which is never published. The individual packages we publish to npm don't have postinstall scripts and so nothing will happen 👍

alexmuller (approved):

jsosa:
I like it!

PR #110, opened 2022-08-05
feat: render full error information

rowanmanning:
This adds all the error information we need to the error info page. It
also introduces styling via Origami and some help text to explain to
engineers what each piece of error information is for.

## Feedback

I'm looking for general feedback on:

  * The approach. We agreed we'd like to avoid a build step or lots
    of additional dependencies. Is this approach still what we want
    now that we have some code to look at?

  * Is there any more useful (and easy to get hold of) information
    that we might want to display on this page?

  * Do we want to try and tackle `error.stack` parsing as part of
    this PR? Or should we aim to get this merged and add as a
    later feature?

## Potential Improvements

We could possibly move each of the `renderX` methods into their
own files, but this might be overkill. For now I split out the layout
as it's mostly static and made the `renderErrorPage` method far
harder to understand.

## How it looks

| With an `OperationalError` | With a non-operational error |
| --- | --- |
| Screengrab | Screengrab |

### DIFF
--- a/packages/middleware-render-error-info/lib/render-error-page.js
+++ b/packages/middleware-render-error-info/lib/render-error-page.js
@@ -25,6 +25,7 @@ const renderLayout = require('./render-layout');
  *     Returns the rendered error page.
  */
 function renderErrorPage({ request, response, serializedError }) {
+	const appName = process.env.SYSTEM_CODE || 'application';
 	return renderLayout({
 		body: `
 			<h1 id="errors">Error information</h1>
@@ -33,7 +34,7 @@ function renderErrorPage({ request, response, serializedError }) {
 			${renderResponse(response)}
 		`,
 		request,
-		response
+		title: escape(`${serializedError.name} in ${appName}`)
 	});
 }
 
--- a/packages/middleware-render-error-info/lib/render-error-page.js
+++ b/packages/middleware-render-error-info/lib/render-error-page.js
@@ -1,0 +1,388 @@
+/**
+ * @module @dotcom-reliability-kit/middleware-render-error-info/lib/render-error-page
+ */
+
+const renderLayout = require('./render-layout');
+
+/**
+ * @typedef {object} ErrorRenderingOptions
+ * @property {import('express').Request} request
+ *     An Express request object.
+ * @property {import('express').Response} response
+ *     An Express response object.
+ * @property {import('@dotcom-reliability-kit/serialize-error').SerializedError} serializedError
+ *     The error to render.
+ */
+
+/**
+ * Render an HTML error info page.
+ *
+ * @access private
+ * @param {ErrorRenderingOptions} options
+ *     Options which impact the rendering of the page.
+ * @returns {string}
+ *     Returns the rendered error page.
+ */
+function renderErrorPage({ request, response, serializedError }) {
+	return renderLayout({
+		body: `
+			<h1 id="errors">Error information</h1>
+			${renderError(serializedError)}
+			${renderRequest(request)}
+			${renderResponse(response)}
+		`,
+		request,
+		response
+	});
+}
+
+/**
+ * Render a serialized error to HTML.
+ *
+ * @access private
+ * @param {import('@dotcom-reliability-kit/serialize-error').SerializedError} error
+ *     The error information to render.
+ * @returns {string}
+ *     Returns the rendered error.
+ */
+function renderError(error) {
+	let warning = '';
+
+	if (!error.stack) {
+		warning = renderWarning({
+			title: 'A non-Error object was thrown.',
+			body: `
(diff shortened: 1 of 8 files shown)


packages/middleware-render-error-info/lib/render-layout.js
```diff
@@ -0,0 +1,127 @@
...
+ *     Returns the rendered error page.
+ */
+function renderLayout({ body, request }) {
+	return `
+		<!DOCTYPE html>
+		<html lang="en">
+			<head>
+				<meta charset="utf-8" />
+				<meta http-equiv="X-UA-Compatible" content="IE=Edge" />
+				<title>Error Info</title>
```
alexmuller, line 61: Might be worth including the name of the app and/or the type of error here?

packages/middleware-render-error-info/lib/render-error-page.js
```diff
@@ -0,0 +1,388 @@
...
+ *     The error information to render.
+ * @returns {string}
+ *     Returns the rendered error.
+ */
+function renderError(error) {
+	let warning = '';
+
+	if (!error.stack) {
+		warning = renderWarning({
+			title: 'A non-Error object was thrown.',
```
alexmuller, line 53: This is cool. Do we want to propose adding [no-throw-literal](https://eslint.org/docs/latest/rules/no-throw-literal) to our standard lint config?

alexmuller (commented):
This is looking really really cool. A couple of thoughts in comments but I don't feel particularly strongly about any of them. Also one more here:

- Do we want to output the system code / name of app on the page for people who are running multiple apps?

In general I think we should get this merged in soon without too much additional work, this PR is big enough! We can have tickets for follow-up.

packages/middleware-render-error-info/lib/render-error-page.js
```diff
@@ -0,0 +1,388 @@
...
+ *     The value to escape.
+ * @returns {string}
+ *     Returns the HTML-escaped value.
+ */
+function escape(value) {
+	return `${value}`
+		.replace(/</g, '&lt;')
+		.replace(/>/g, '&gt;')
+		.replace(/"/g, '&quot;');
+}
```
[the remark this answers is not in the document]
rowanmanning: Yeah totally, it _is_ a security risk and it made me feel really iffy writing it, however I think the attack vector isn't one that someone would try and exploit? So the only ways I can think of this being a vector (please let me know if I'm wrong) are:

  * during a supply chain attack, where an error thrown by one of our app dependencies includes malicious information that embeds a script or something in our error page. This feels like a round-the-houses way to gain privileged access considering that the dependency has near-full access to the user's machine if they have code running in a Node.js process

  * If the request headers contain a similar exploit (in which case the user's browser is probably compromised)

Saying that, I think maybe moving to a third-party package as part of this makes sense. [I'm gonna have a go at switching to this](https://github.com/fb55/entities).

packages/middleware-render-error-info/lib/render-error-page.js
rowanmanning: Yes let's propose that. We do mention using this and [`prefer-promise-reject-errors`](https://eslint.org/docs/latest/rules/prefer-promise-reject-errors) in the [linked documentation](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/getting-started/throwing-errors.md#using-error-objects) too

andygout (approved):
Nice! 👍 

> The approach. We agreed we'd like to avoid a build step or lots
of additional dependencies. Is this approach still what we want
now that we have some code to look at?

I think you've written the code in a way that keeps it easy enough to add and remove properties we want to display, which I would imagine will be the extent of how we want to modify the code going forwards. I think having additional build steps and dependencies in the context of the consuming app creates the potential for ongoing friction and so my vote would still be for this approach.

> Is there any more useful (and easy to get hold of) information
that we might want to display on this page?

Maybe some basic stuff that can sometimes create gotchas:
- commit hash
- commit name
- branch name
- Node.js version
- flag states (though this would occupy a lot of space so maybe a link to toggler.ft.com reminding people that the state of their flags might affect the app if running via the router)
- user licences (e.g. having an FT.com Premium (P2) licence will affect what the user sees)

(If any of these are deemed useful then adding them on a separate branch would make sense as there is already lots in this PR)

> Do we want to try and tackle error.stack parsing as part of
this PR? Or should we aim to get this merged and add as a
later feature?

I think a later feature as there is already a huge amount of useful information displayed in this PR alone.

> We could possibly move each of the renderX methods into their
own files, but this might be overkill. For now I split out the layout
as it's mostly static and made the renderErrorPage method far
harder to understand.

I have no strong feeling for either at this stage.

rowanmanning:
I think these are all great suggestions, and agree we should add more info in a later PR 👍

PR #116, opened 2022-08-10
Refactor the tests

rowanmanning:
I wasn't quite happy with our tests so I decided to make a few changes, most centre around refocusing the `examples` directory on sample code rather than trying to piggy-back on it for end-to-end tests. This comes as a few separate commits which may be easier to review separately.

Here are the changes that have been made and why:

## Purpose of examples

The purpose of the `examples` folder wasn't clear, and we were adding in code specifically used for testing rather than showing users how to use Reliability Kit. I think this was a smell. It also meant we had to explicitly drop our 100% coverage requirement because it's not realistic for an example which includes a lot of boilerplate.

I've deleted the tests in this directory. I'm confident doing this because actually all we were testing was:

  * The `serialize-error` and `serialize-request` packages, which already have good tests and 100% coverage.
    Their unit tests are enough

  * That our mock JSON error render was working, and this isn't something we're shipping or expecting users to add

  * We can add _actual_ integration tests for each package that needs them, as outline below

## Integration tests in packages

I've updated all the package unit tests to be in a `test/unit` folder rather than just under `test`. This paves the way for adding integration tests on a per-package basis, which I think we're lacking in a few cases. To illustrate how we can write better integration tests for modules I added a new test for the logging middleware.

This new test is far better than the original Express test, which still mocked `@financial-times/n-logger` so it could verify that logging was being performed. The new test spins up an Express app in a child process and collects the actual logs which end up in `stderr`. This means we now have a true end-to-end test which will fail if there's an unexpected API change in n-logger or n-express.


andygout (approved):
**praise:**
Thanks for breaking the work into individual commits - made it really easy to follow.

**question:**
I've not seen `child_process` used for an end-to-end test before - is that a Node.js native function? (The npm registry [`child_process`](https://www.npmjs.com/package/child_process) package is a security holding package so it seems that way.)

What are the benefits of using this over something like `supertest` or `chai-http` (in addition to having one less dependency)? Would the tests you've written even be possible with those packages (they are modifying the child process in a way I have not seen done in tests before)?

**comment:**
I think there are varying definitions of integration and end-to-end tests, with some definitions conflating the two.

I have gradually formed these definitions, as I feel there is a distinction between integration and end-to-end tests (or rather I have found that there is sometimes a requirement for what I have described as integration tests - maybe those should have another name altogether):
- **unit tests:** tests a file/module in complete isolation; all imports from outside the module are stubbed/mocked
- **integration tests:** tests the interaction of multiple files/modules in conjunction with one another; the starting point for the test will still be the invocation of one of the modules' functions; dependencies and resources external to the app (e.g. packages imported from the npm registry, databases) will still be stubbed/mocked
- **end-to-end tests:** test the full scope of a request being made to an app's entry point and the resultant response; no dependencies or resources are stubbed/mocked

Would be keen to hear thoughts on this comment.

rowanmanning:
Thanks Andy, I'll attempt to answer these.

## On `child_process`

> I've not seen `child_process` used for an end-to-end test before - is that a Node.js native function?

Yep [`child_process`](https://nodejs.org/api/child_process.html) is built-in to Node.js. Many other modules which run commands from Node.js use this under the hood.

> What are the benefits of using this over something like `supertest` or `chai-http` (in addition to having one less dependency)? Would the tests you've written even be possible with those packages (they are modifying the child process in a way I have not seen done in tests before)?

I did originally write the test like this, but it made it near-impossible to inspect the `console` output in a way that didn't involve mocking `@financial-times/n-logger`. The reason I'm using `child_process` here is because it means our Express app runs in a completely different Node.js process and we can capture the logs as they reach `process.stderr`. That's what we're doing in this code here:

```js
child.stderr.on('data', (chunk) => {
    stderr += chunk.toString();
});
```

This means that the tests are verifying that the logs reach the console (and thus Heroku's log drains), and anything that happens across the stack of dependencies which change the way this works will cause the test to fail. I don't think it'd be possible to ensure this with Supertest or Chai HTTP.

## On definitions

> I think there are varying definitions of integration and end-to-end tests, with some definitions conflating the two.

Yeah I think this actually maps to my understanding of these terms too, in which case `integration` might be the wrong name for these tests because it does ensure that things work fully end-to-end with no mocking. I named it integration out of habit because I rarely end up writing end-to-end tests for library code. Are you happy for me to change the path to `test/end-to-end`?

andygout:
> Are you happy for me to change the path to `test/end-to-end`?

Yes 😄

PR #122, opened 2022-08-11
feat: conceal sensitive values in middleware-render-error-info display

andygout:
FTDCS-274

Jira card: https://financialtimes.atlassian.net/jira/software/c/projects/FTDCS/boards/1426?modal=detail&selectedIssue=FTDCS-274

> We realised after building it that it’s very easy in [middleware-render-error-info](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/middleware-render-error-info#readme) to forget that it’s displaying sensitive information. It’s reasonable to think that someone might share a screen-grab of this page while debugging something and that’s an issue if we display things like API keys or cookie values in the clear.
> I think there are a few options with different levels of difficulty:
> - Obfuscate the value for these fields (we’d need to decide on a list, and I’m sure we could miss some sensitive headers). We could do this by x-ing out half the data?
> - Don’t show the value for these fields at all, instead display a note saying that it’s hidden for security reasons
> - Instead of showing the values for these fields by default, put them in a [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element which is collapsed by default and include a note not to screen-grab while they’re expanded.
>
> I’m happy with any of these options. Interested if there are others, or if we think this isn’t a big enough risk to address with code.

I opted for the second option. The first presented the complexity of calculating how much of a given string to obscure (you could hypothetically have short but sensitive value where obscuring half would still be insufficient) and the third still had the potential for users to accidentally capture sensitive details in a screenshot.

I applied the concealment to both the request query params and headers as there is always the possibility, however ill-advised, an `x-api-key` is passed in the query params.

I have also lowercased the values before comparison to account for varying casing of values, e.g. `x-api-key`, `X-Api-Key` are both used because header names are not case sensitive.

Screenshot 2022-08-11 at 16 58 39

### DIFF
--- a/packages/middleware-render-error-info/lib/render-error-page.js
+++ b/packages/middleware-render-error-info/lib/render-error-page.js
@@ -5,6 +5,17 @@
 const entities = require('entities');
 const renderLayout = require('./render-layout');
 
+const X_API_KEY_REQUEST_PROPERTY_NAME = 'x-api-key';
+const COOKIE_REQUEST_PROPERTY_NAME = 'cookie';
+
+const CONCEALED_REQUEST_PROPERTY_NAMES = new Set([
+	X_API_KEY_REQUEST_PROPERTY_NAME,
+	COOKIE_REQUEST_PROPERTY_NAME
+]);
+
+const CONCEALED_VALUE_MESSAGE =
+	'** This value is likely sensitive and should be examined via other tools **';
+
 /**
  * @typedef {object} ErrorRenderingOptions
  * @property {import('express').Request} request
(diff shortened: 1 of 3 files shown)


rowanmanning (dismissed):
This looks good to me 👍 nice work. I think you're right, this is the easiest way to tackle the problem and we can add new restricted properties if we stumble across them later

rowanmanning (changes requested):
Code looks all great to me, the only change is that this shouldn't be a `chore` commit as it won't result in a release being tagged up. I think this is either a `feat` or a `fix`? Maybe slightly leaning towards feature but could go either way 🤷

rowanmanning (approved):
🎉

andygout:
I have changed the commit name and PR description from `chore` to `feat` but cannot change the branch name and think creating an entirely new branch is overkill if it does not factor into the release decision-making process?

PR #123, opened 2022-08-12
docs: correct a mistake in commit type prefixes

rowanmanning:
The `docs` commit prefix _does_ trigger a patch release
which I think is what we want (docs changes should make
it to npm right?)

[Evidence: see `docs` isn't hidden in the config](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/release-please-config.json#L4-L23).


alexmuller (approved):
That makes sense - although I would expect 90% of docs changes to come with a `feat` and only rarely get fixed standalone.

PR #125, opened 2022-08-12
chore: set a time and timezone for Dependabot

rowanmanning:
I think we're getting these at night, I wonder if it's
better if it runs first thing – I often use the first half
hour of the day for chores like dependency bumps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,34 +1,58 @@
 version: 2
 updates:
+
   - package-ecosystem: "npm"
     directory: "/"
     schedule:
       interval: "daily"
+      time: "09:00"
+      timezone: "Europe/London"
+
   - package-ecosystem: "npm"
     directory: "/logos"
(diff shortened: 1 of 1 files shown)


andygout (commented):
I might be missing something but surely if they run during the night then they are still ready for you to address in the first hour of your working day?

rowanmanning:
True but you get the Slack and GitHub notifications late at night. I have them turned off for Slack but this will stop me from being tempted to check GitHub 😁 (I also have my GitHub PR notifications open in a personal tab for my own open source projects so that I can merge my own dependabot PRs)

andygout (approved):
Oh, I see. You are speaking to somebody who turns their laptop and phone off at around 8pm and turns them on again in the morning.

rowanmanning:
Sounds healthier tbh, I still maintain enough OSS that I occasionally need to do some maintenance!

PR #130, opened 2022-08-15
feat: add new more specific error classes

rowanmanning:
This adds some new error classes to the `errors` package which should
help apps provide more detailed error information with relatively little
effort. It adds in:

  * `DataStoreError`
  * `UpstreamServiceError`
  * `UserInputError`

I based these on [the needs of next-home-page](https://github.com/Financial-Times/next-home-page/blob/main/apps/home-page/lib/errors.ts), one of our more
recently-built apps which does error handling quite well. I also did
some digging across all of [the FT's repos](https://cs.github.com/?q=extends+Error+%28language%3AJavaScript+OR+language%3ATypeScript%29&scope=org%3AFinancial-Times&scopeName=Financial-Times) to get an idea of what error
classes our apps are defining themselves.

Resolves FTDCS-285.

### DIFF
--- a/packages/errors/lib/data-store-error.js
+++ b/packages/errors/lib/data-store-error.js
@@ -1,0 +1,29 @@
+/**
+ * @module @dotcom-reliability-kit/errors/lib/data-store-error
+ */
+
+const OperationalError = require('./operational-error');
+
+/**
+ * Class representing an error in an application's data store.
+ */
+class DataStoreError extends OperationalError {
+	/**
+	 * @readonly
+	 * @access public
+	 * @type {string}
+	 */
+	name = 'DataStoreError';
+
+	/**
+	 * Create a data store error.
+	 *
+	 * @param {(string | OperationalError.OperationalErrorData & Record<string, any>)} [data = {}]
+	 *     The error message if it's a string, or full error information if an object.
+	 */
+	constructor(data = {}) {
+		super(data);
+	}
+}
+
+module.exports = DataStoreError;
--- a/docs/getting-started/handling-errors.md
+++ b/docs/getting-started/handling-errors.md
@@ -129,7 +129,7 @@ In this scenario we have the following possible errors which we know can make it
 
   * The fruit API does not have a fruit matching `request.params.name` and so the client throws an error with a `statusCode` property set to `404`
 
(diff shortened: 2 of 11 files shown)


packages/errors/lib/data-store-error.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Class representing an error in an application's data store.
+ */
+class DataStoreError extends OperationalError {
+	/**
+	 * @readonly
+	 * @access public
+	 * @type {string}
+	 */
+	name = 'DataStoreError';
```
andygout, line 16: **question:** I'm still getting my head around JSDocs. In this code block it looks like `name`  is not commented out but neither is it valid code as it is not preceded with `var`, `let`, or`const`.

I'm also unsure in this context what `name` refers to as it does not seem to be referenced elsewhere in this file.

Can you help me understand pls?

andygout (approved):
👍 with one question for my education 🙏

packages/errors/lib/data-store-error.js
rowanmanning: Ooh this is a [public class field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields), it's a relatively new (Node.js 14+) feature of JavaScript which allows you to define instance properties on the class definition rather than having to set `this.whatever` in the constructor.

So in this `DataStoreError` class it means that:

```js
const error = new DataStoreError('Something went wrong');
error.name === 'DataStoreError'; // true
```

Even more exciting than this is the existence of [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) (not used here), which are only accessible by methods on the instance of a class so we have actual private properties in classes now.

rowanmanning: These are _kind_ of equivalent (with a few minor differences explained in the Mozilla docs):

```js
// Class fields
class Example {
    myProperty = 123;
}
```

```js
// Constructor
class Example {
    constructor() {
        this.myProperty = 123;
    }
}
```

andygout: Amazing - thanks for the explanation 🙇

alexmuller (approved):
This looks good - tests are all doing the right thing and running for me.

Only changes are to the errors package which I think means the `docs/` changes in this PR are fine. They will go out with that package only?

PR #131, opened 2022-08-15
docs: clarify that x-request-id is auto-logged

rowanmanning:
We weren't being very clear that the `X-Request-ID` header is already
automatically logged under the `request.id` property. I've attempted to
clarify this in the documentation.

If this is still not clear then another option would be to ignore this
header if it's passed in as part of the `includeHeaders` option.

@gyss I updated this documentation based on the inclusion of the
`X-Request-ID` header [here](https://github.com/Financial-Times/dotcom-pages/pull/5#discussion_r945760738). Do you think it's clearer now / you would
have known there was no need to include it?

### DIFF
--- a/packages/log-error/README.md
+++ b/packages/log-error/README.md
@@ -49,7 +49,9 @@ This will automatically [serialize error objects](https://github.com/Financial-T
     message: 'Error: something went wrong',
 
     error: {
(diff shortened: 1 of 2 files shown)


gyss (approved):

gyss:
Looks good! thanks @rowanmanning

andygout (approved):

PR #132, opened 2022-08-15
chore: start testing in Node.js 18

rowanmanning:
Node.js 18 isn't in active Long-Term-Support yet, but I think it's worth
starting to test against it as it becomes an LTS version of Node.js in
October 2022. This ensures that once we're ready to move apps to v18 our
modules will work as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ references:
       - image: cimg/node:<< parameters.node-version >>
     parameters:
       node-version:
-        default: "16.15"
+        default: "16.15" # We default to the highest active LTS
         type: string
 
   workspace_root: &workspace_root ~/project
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -151,25 +151,27 @@ workflows:
...
-              node-version: [ "16.15", "14.19" ]
+              node-version: [ "18.7", "16.15", "14.19" ]
       - lint:
           requires:
             - build-v<< matrix.node-version >>
           name: lint-v<< matrix.node-version >>
           matrix:
             parameters:
-              node-version: [ "16.15", "14.19" ]
+              node-version: [ "18.7", "16.15", "14.19" ]
```
andygout, line 168: Does the steps being run for each Node.js version run in tandem or successively?

I'm trying to gauge if this might potentially slow down our CI process.

If they run in tandem (which I think they do based on how a typical [`build-test` flow](https://github.com/Financial-Times/dotcom-reliability-kit/runs/7838898914) plays out), might running the same steps for an increasing number of Node.js versions somehow else cause bottlenecks (e.g. the number of simultaneous CircleCI containers our plan provides)?

I am raising these points based on some previous Platforms work to try and avoid/reduce the length of CircleCI queues.

andygout (approved):
👍 

I have asked one non-blocking question.

.circleci/config.yml
rowanmanning: They run in tandem yep, and yes that could potentially cause bottlenecks if our CircleCI plan doesn't quite cater for it. _however_ I think:

1. testing against multiple stable Node.js versions is something we need and our CI tooling needs to handle that. If it's an issue we could either try and get more capacity or move to GitHub actions which would be free (and not impact our quota) as this is an open source repo
2. [Node.js 14 end-of-life is April 2023](https://nodejs.org/en/about/releases/) so we'll need increased capacity between October 2022 and April 2023 at which point we can drop Node.js 14 support

rowanmanning: Another alternative is we leave this open as a long-running PR and just periodically rebase it and run the tests in Node.js 18, then we merge in October to reduce the window where we need increased capacity

andygout: October will come soon enough so I have no issues with merging now.

PR #138, opened 2022-08-17
fix: default UpstreamServiceError to HTTP 502

alexmuller:
We agreed that a 502 is a more sensible default than a 500

- https://github.com/Financial-Times/dotcom-debug-user/pull/139#discussion_r946863853
- [Slack](https://financialtimes.slack.com/archives/CPNKK12S1/p1660661442019349)

### DIFF
--- a/packages/errors/lib/upstream-service-error.js
+++ b/packages/errors/lib/upstream-service-error.js
@@ -29,6 +29,17 @@ class UpstreamServiceError extends HttpError {
 		if (typeof data === 'number') {
 			data = { statusCode: data };
 		}
+
+		// Make sure that we don't modify the original data object
+		// by shallow-cloning it
+		data = { ...data };
+
+		// Default the status code
+		data.statusCode =
+			typeof data.statusCode === 'number'
+				? UpstreamServiceError.normalizeErrorStatusCode(data.statusCode)
+				: 502;
+
 		super(data);
 	}
 }
--- a/docs/getting-started/handling-errors.md
+++ b/docs/getting-started/handling-errors.md
@@ -161,13 +161,12 @@ app.get('/fruit/:name', async (request, response, next) => {
         }
 
         // If the fruit API was down, let's throw an error which
-        // better describes what's happened. We switch the status
-        // code to 502 because that matches what's happening – we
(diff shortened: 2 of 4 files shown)


rowanmanning (approved):
Looks perfect 👍 well done spotting the "handling errors" guide, I already forgot that we were setting this manually there!

packages/errors/lib/upstream-service-error.js
```diff
@@ -29,6 +29,17 @@ class UpstreamServiceError extends HttpError {
...
 		}
+
+		// Make sure that we don't modify the original data object
+		// by shallow-cloning it
+		data = { ...data };
+
+		// Default the status code
+		data.statusCode =
+			typeof data.statusCode === 'number'
+				? UpstreamServiceError.normalizeErrorStatusCode(data.statusCode)
```
andygout, line 40: **question:** So if an upstream service returns a 456 error then we set that as the `UpstreamServiceError`'s status code but if no status code is given then it is set to 502 to convey that something has gone wrong upstream?

andygout (approved):
👍 I have asked one non-blocking question.

packages/errors/lib/upstream-service-error.js
rowanmanning: Nope, if an upstream service returns a `456` error then we still send `502`. The conditional here is looking at the `data` object which is supplied by the user, so it's possible to override the `502` but you have to do it explicitly:

```js
throw new UpstreamServiceError({
    // This is `data`
    message: 'Hello',
    statusCode: 456
});
```

I think it's still best to leave the `502` in place in most cases though, and instead pass through a `cause`:

```
throw new UpstreamServiceError({
    message: 'Hello',
    cause: originalFetchError
});
```

because then it'll serialize in a way that's super useful in the logs:

```js
{
    error: {
        name: 'UpstreamServiceError',
        statusCode: 502,
        cause: {
            name: 'FetchError',
            statusCode: 456
        }
    }
}
```

andygout: Ah, I see - so it's down to advised usage of `UpstreamServiceError` 👍

PR #143, opened 2022-08-22
fix: refactor to centralise environment variables

rowanmanning:
This tucks away nearly all of our environment variable access into a new
module named app-info. This means we're no longer repeating some logic
across the other modules and can remove a lot of tests (despite adding a
new package and a bunch of documentation, this decreases the overall
lines of code by ~60).

This also adds a feature: we're now not only looking at the
`SYSTEM_CODE` environment variable, we're also attempting to load the
`package.json` file in the current working directory and normalising the
name to get a default system code. This nearly* matches the behaviour in
n-express.

A benefit of this work is that if we need to use the system code in
other places (which I think is likely, e.g. if we wrap any of the
tooling for OpenTelemetry). We could also replace usage of all these
environment variables across our other libraries and apps so that
there's a single source of truth for getting the app's system code.

---

*The system code sanitisation only removes a prefix of `ft-` which
doesn't quite match the [n-express behaviour](https://github.com/Financial-Times/n-express/blob/main/src/lib/normalize-name.js). There are two things in
this that I don't think we need any more:

  1. The n-express library is designed to normalise a _name_ and not a
     system code, so it removes the `next-` prefix as well. We don't
     want to do this as many of our system codes _do_ begin with `next-`

  2. The n-express library removes a suffix of `-vXXX` (`X` being any
     number). From what I can tell, at some point we intended to version
     package names but this is not used in any of our current apps.
     https://github.com/Financial-Times/n-express/commit/a598c601040aabbc35e06efc1cd0713cf692d1b5

### DIFF
--- a/packages/app-info/README.md
+++ b/packages/app-info/README.md
@@ -1,0 +1,70 @@
+
+## @dotcom-reliability-kit/app-info
+
+A utility to get application information (e.g. the system code) in a consistent way. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
+
+  * [Usage](#usage)
+    * [`appInfo.commitHash`](#appinfocommithash)
+    * [`appInfo.environment`](#appinfoenvironment)
+    * [`appInfo.region`](#appinforegion)
+    * [`appInfo.releaseDate`](#appinforeleasedate)
+    * [`appInfo.releaseVersion`](#appinforeleaseversion)
+    * [`appInfo.systemCode`](#appinfosystemcode)
+  * [Contributing](#contributing)
+  * [License](#license)
+
+
+## Usage
+
+Install `@dotcom-reliability-kit/app-info` as a dependency:
+
+```bash
+npm install --save @dotcom-reliability-kit/app-info
+```
+
+Include in your code:
+
+```js
+import appInfo from '@dotcom-reliability-kit/app-info';
+// or
+const appInfo = require('@dotcom-reliability-kit/app-info');
+```
+
+The `appInfo` object has several properties which can be used to access application information.
+
+### `appInfo.commitHash`
+
+Get the commit hash that the application last deployed. This will be a string (if `process.env.HEROKU_SLUG_COMMIT` is defined) or `null` otherwise. This relies on the [Heroku Dyno metadata labs feature](https://devcenter.heroku.com/articles/dyno-metadata) and will not be present in local development.
+
+### `appInfo.environment`
+
+Get the application environment, normally either `development` or `production`. This will be a string, using `process.env.NODE_ENV` and defaulting to `development`.
+
+### `appInfo.region`
+
+Get the application Heroku region. This will be a string (if `process.env.REGION` is defined) or `null` otherwise.
+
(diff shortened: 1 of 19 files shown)


packages/app-info/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Get the commit hash that the application last deployed. This will be a string (if `process.env.HEROKU_SLUG_COMMIT` is defined) or `null` otherwise. This relies on the [Heroku Dyno metadata labs feature](https://devcenter.heroku.com/articles/dyno-metadata) and will not be present in local development.
+
+### `appInfo.environment`
+
+Get the application environment, normally either `development` or `production`. This will be a string, using `process.env.NODE_ENV` and defaulting to `development`.
+
+### `appInfo.region`
+
+Get the application Heroku region. This will be a string (if `process.env.REGION` is defined) or `null` otherwise.
```
andygout, line 45: Is it worth stating that this will not be present in local development or is that implicit?

.github/dependabot.yml
```diff
@@ -8,6 +8,13 @@ updates:
       time: "09:00"
       timezone: "Europe/London"
 
+  - package-ecosystem: "npm"
+    directory: "/examples/express"
+    schedule:
+      interval: "daily"
+      time: "09:00"
+      timezone: "Europe/London"
```
andygout, line 16: **question:** Am I right in thinking that the PR is not dependent on this change?

andygout (approved):
I have asked a couple of non-blocking questions.

.github/dependabot.yml
rowanmanning: Yes you are, oops! I'll extract this

packages/app-info/README.md
rowanmanning: I _think_ it's implicit and not always true, because if a team sets a `REGION` environment variable in their `development` Vault config then it will be present?

.github/dependabot.yml
alexmuller: @rowanmanning But also should there be a dependabot change on this PR for the `app-info` package?

alexmuller: Oh! Or not because it has no dependencies!

rowanmanning: Yep I feel like _maybe_ that'll trip us up if this package does ever have dependencies, but I think it's quite unlikely to ever need any?

alexmuller (approved):
This looks good - I think I understand it and I think Reliability Kit is the right place for it. Although it could also live in a completely separate package. It's almost certainly a good move to abstract these environment variable names to remove implementation details like the name of our current hosting provider.

At some point in the future it may end up being useful to parse the release date from a string.

rowanmanning:
Thanks!

1. I did almost parse the date from a string, but then decided as we don't actually need that right now, and it gets converted back to a string anyway in the logs (the only place we currently use it). Definitely may need a date later though

2. I also considered another package, but this should be super easy to extract from Reliability Kit if we decide to do that later :) right now it's easier to keep it inside because all the packages that depend on it _within_ Reliability kit will get automatic version bumps when we release a new version of it

PR #211, opened 2022-10-05
feat: add an experimental logger package

rowanmanning:
> **Note**
> This is an experimental piece of work which may never make it into production. There's no agreed timeline for migrating things to this new logger - we want to be careful about introducing too many Reliability Kit migrations at once.

This adds a new experimental package which aims to be a near drop-in replacement for n-logger and a replacement for n-serverless-logger with a little work. I spent a lot of time working on a suite of end-to-end tests to ensure that it's as compatible as possible.

I think the next step will be to release a beta version of this package (v0.1.0) so that we can test it in some non-critical applications.

I tried to approach the build of this in the same way I would a package that'd definitely go into production, because writing tests and documentation after the fact is always difficult. Despite it looking relatively polished, this can be picked apart and rethought as much as we need to.

Reviewing
---------

This review can happen more slowly than others because we're adding something experimental without having planned the approach as a team. I worked on this in 10% time because I was interested in the problem space and wanted a downtime project where I could focus a lot on testing.

I think it'd also be good to get some feedback from people outside of the CP Reliability team because they're closer to how their apps do logging.

Areas to focus on are:

  * Are the migration guides clear enough. Using these alone, do you think you could migrate an application away from n-logger or n-serverless-logger?

  * Do you agree with the deprecation decisions and where there _is_ a breaking change between libraries?

  * Are the end-to-end tests thorough enough? I think these are the most important part of the library because they help us be sure that we're as compatible as possible. My intention is that if we find use cases for n-logger which we're _not_ compatible with then we should add a new test case.

Some fun stats
--------------

I'm quite pleased already with the difference between this and n-logger plus n-serverless-logger (both of which it replaces) in terms of some code metrics. These aren't perfect measurements but they do give an idea of what the maintenance difference is.

stat | n-logger | n-serverless-logger | new package | diff
-----|----------|---------------------|-------------|-----
lines (source) | `230` | `221` | `177` | `-274`
lines (tests) | `612` | `140` | `1292` | `+540`
lines (comments) | `47` | `33` | `378` | `+298`
ratio (source:tests) | `1:3` | `1:2` | `1:7` | -
ratio (source:comment) | `5:1` | `7:1` | `1:2` | -

Next steps
----------

  1. Release a beta version of this package and test it in some non-critical applications that this team owns

  2. Provide a migration path for n-mask-logger, because I think this work is only worth doing if we can deprecate all three logging modules. I have some ideas for how we could do this but didn't want this branch to get any bigger before we have some more eyes on it

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -62,6 +62,16 @@ updates:
       prefix-development: "chore:"
 
   - package-ecosystem: "npm"
+    directory: "/packages/logger"
+    schedule:
+      interval: "daily"
+      time: "09:00"
+      timezone: "Europe/London"
+    commit-message:
+      prefix: "fix:"
+      prefix-development: "chore:"
+
+  - package-ecosystem: "npm"
     directory: "/packages/middleware-log-errors"
     schedule:
       interval: "daily"
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -3,6 +3,7 @@
   "packages/crash-handler": "0.1.0",
   "packages/errors": "1.2.6",
   "packages/log-error": "1.3.8",
+  "packages/logger": "0.0.0",
   "packages/middleware-log-errors": "1.2.9",
   "packages/middleware-render-error-info": "1.1.7",
   "packages/serialize-error": "1.1.4",
--- a/jsconfig.build.json
+++ b/jsconfig.build.json
@@ -7,9 +7,6 @@
 		"noEmit": false,
 		"removeComments": true
 	},
-	"exclude": [
-		"**/test/**/*"
-	],
 	"include": [
 		"packages/**/*.js"
 	]
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -18,7 +18,8 @@
 	"exclude": [
 		"coverage",
 		"node_modules",
-		"**/*.spec.js"
+		"**/*.spec.js",
+		"**/test/**/*"
 	],
 	"include": [
 		"**/*.js"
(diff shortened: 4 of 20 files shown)


wheresrhys:
A couple of thoughts
1. Logging of error objects - This always confused me (possibly because of not being documented clearly enough e.g. should it be in a property called `error`, `err`, passed as a second parameter, destructured at the top level? Woudl be good to get some clear guidance over which patterns are advised and what the behaviour is. There's one example in your docs, but it's not clear if this is the only way to log an error well (compare with how `.info(obj, message)`, `.info(message, obj)`, `.info({...obj, message})` all do the same thing (or do they?? - maybe that needs to be documented clearly too? And my point is more about how additional parameters get absorbed into the final logged object)
2. Transport patterns for logging in lambda - As far as I know the pattern used by n-serverlesslogger is not one advised by cloudenablement (at least, it wasn't a few years ago). lambda-logger did follow a log drain pattern advised by CE back in the day, but even that may be a bit dated now. Might be worth talking to CE for guidance (if you haven't already done so)

rowanmanning:
Thanks for the feedback @wheresrhys! Agree on the docs, [how's this](https://github.com/Financial-Times/dotcom-reliability-kit/blob/add-experimental-logging-package/packages/logger/README.md#log-data-serialization)?

wheresrhys:
Yep - excellent!

alexmuller (approved):
🟡 The emojis in the migration guide are incredibly helpful

I've not looked at this in _huge huge_ detail but I think it's very cool.

PR #229, opened 2022-10-11
Publish TypeScript type definitions

rowanmanning:
This ensures that when we publish to npm, we generate and include TypeScript type declarations (`.d.ts`) with our modules. These are generated from the JSDoc and map 1:1 with what we've written. It's annoying that we had to do this but I think it makes the most sense when you consider the alternatives.

## Why

@jamesr101 spotted today that with a full TypeScript-based projects, he was getting type errors and no type hinting in VSCode. I tried a basic TypeScript project and it works fine unless you start configuring with a `tsconfig.json` file. Once you add that in, type hinting in VSCode seems to break. This is why we haven't spotted this before – our initial testing was too basic.

I found out that this is a known issue and that JavaScript-based projects which use JSDoc types are not actually usable by TypeScript projects without some extra configuration:

  * https://github.com/microsoft/TypeScript/issues/19145
  * https://github.com/microsoft/TypeScript/issues/29824

I decided that the best solution would be to generate type declarations just before publishing to npm so that our types are picked up in projects that depend on Reliability Kit.

## Alternatives

I considered a few other ways we could address this:

1. **Move this project fully over to TypeScript:** I think this would be an annoying and large job. I also stand by the fact that it's nice that whatever we write is exactly what's running on our servers. If we want to explore this then I think it's a separate piece of work and not something we should take on in a rush to fix the immediate issue

2. **Require that TypeScript-based projects configure themselves to suit Reliability Kit:** it's possible to get around this by adding `"maxNodeModuleJsDepth": 2` to your `.tsconfig.json` file. This tells TypeScript to look two levels deep in `node_modules` for JSDoc-based type declarations. Something feels off about requiring specific configurations just to use Reliability Kit and I don't want to add any barriers to adoption. I have no idea if adding this configuration will cause more pain in projects as third-party JSDoc would be picked up too.

## Review notes

This seems pretty large, but really most of the changes are tweaks to the JSDoc to make sure that it compiles to the correct/expected TypeScript type declarations. Each commit is a distinct piece of work and the more complex commits have a description which explains why the work needed doing.

To test the generated TypeScript files locally, you can run:

```
npm run build
```

Then open a module's `lib` folder and you should see matching `.d.ts` files.

The testing I've done locally is generating these files, symlinking my local Reliability Kit into different Typescript-based CP apps, and running the build to make sure there are no type errors. I also verified that type-hinting in VS Code is working.

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -3,6 +3,7 @@
 		"allowJs": true,
 		"checkJs": true,
 		"module": "commonjs",
+		"noEmit": true,
 		"resolveJsonModule": true,
 		"strictBindCallApply": true,
 		"strictFunctionTypes": true,
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,6 +1,5 @@
 {
 	"compilerOptions": {
-		"allowJs": true,
 		"checkJs": true,
 		"module": "commonjs",
 		"noEmit": true,
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,9 @@ jobs:
           command: echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" >
             ${HOME}/.npmrc
       - run:
+          name: Build packages
+          command: npm run build
+      - run:
           name: NPM publish
           command: ./scripts/circleci-publish.sh
 
@@ -138,6 +141,9 @@ jobs:
           name: Set npm auth token
           command: echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" >
             ${HOME}/.npmrc
+      - run:
+          name: Build packages
+          command: npm run build
       - run:
           name: NPM publish
           command: ./scripts/circleci-publish.sh --tag=prerelease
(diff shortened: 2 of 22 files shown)


alexmuller (commented):
I think this all makes sense although I am by no means the best person to review it - my knowledge of TypeScript is almost 0. I can build locally and I can see that the build will fail if we include any bad types.

jsconfig.json
```diff
@@ -3,6 +3,7 @@
 		"allowJs": true,
```
oliverturner, line 3:
```suggestion
```
Nit [non-blocking]: `"allowJs": true` is implicit in the use of `jsconfig.json`

oliverturner: Not to detract from an otherwise excellent-looking PR!

jamesr101 (approved):
Excellent work, that looks good to me

jsconfig.json
rowanmanning: Good spot! Thanks :)

andygout (approved):
Thanks for breaking the steps down into individual commits as made it infinitely easier to digest what the changes are doing.

PR #233, opened 2022-10-12
fix: include TypeScript types in the build

rowanmanning:
This is annoying. I forgot that `.npmignore` also uses `.gitignore` rules and so our TypeScript files were being ignored.


andygout (approved):
Non-blocking question: what does the exclamation mark at the start of the pattern do?

PR #241, opened 2022-10-13
Add stricter type checking

rowanmanning:
This adds in some stricter type checking which is easy to add now but more difficult to add later. Most of these are properties which can be added without having to change any Reliability Kit code, the exception being `noImplicitOverride` (see below). I think it's worth adding these in so that we catch these kinds of issues if they're ever introduced.

**`noImplicitOverride`:** This option requires you to acknowledge when you're overriding a property or method on a parent class. This ensures that TypeScript will error if we ever change methods or properties on a base class but miss any extending classes. It requires that we add an `@override` tag to extended class members and I've added these.

[Documentation for all of these options is here](https://www.typescriptlang.org/tsconfig).

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -1,8 +1,11 @@
 {
 	"compilerOptions": {
+		"alwaysStrict": true,
 		"checkJs": true,
+		"exactOptionalPropertyTypes": true,
 		"module": "commonjs",
 		"noEmit": true,
+		"noImplicitThis": true,
 		"resolveJsonModule": true,
 		"strictBindCallApply": true,
 		"strictFunctionTypes": true,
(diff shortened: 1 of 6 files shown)


andygout (commented):
So I am clear, TypeScript will only error if we override a property on an extension of that class with specifying `@override`?

And does override mean changing the `@type` or would adjusting other properties (e.g. `@readonly`, `@public`) also be considered an override?

rowanmanning:
`@override` is required if you declare _anything_ on a child class which matches a name on a parent class. It doesn't matter if you make modifications to the signature of that property/method, you still need to explicitly say you're overriding. I don't _think_ that changing `@readonly` or `@public` is actually valid anyway under TypeScript (even without this change)

andygout (approved):

PR #254, opened 2022-10-20
feat: add an uncaught exception handler

### LINKED ISSUE #225: Write an uncaught exception handler package
As part of our [investigation into Sentry usage](https://financialtimes.atlassian.net/l/cp/x3M1jCkb), we may end up building an experimental package to replace Sentry's uncaught exception handler.

We currently rely completely on [n-raven](https://github.com/Financial-Times/n-raven) to register an [uncaught exception handler](https://nodejs.org/api/process.html#event-uncaughtexception). If we're considering making Sentry optional for our apps, then we need a simpler replacement module to handle uncaught exceptions, otherwise these errors will crash apps without any visibility in our monitoring.

## Ideal solution

Uncaught exceptions should go to Splunk, the same as other errors. This requires us to bind our own handler to the Node.js process `uncaughtException` event. This should be done explicitly in each app so that they can configure it and so that engineers in our teams can see and understand the code that's running.

Name TBC, but I quite like `crash-handler` as a package name.

I think the code would be relatively minimal, as we already have a `logUnhandledError` method which was built with this task in mind. I wrote this while experimenting as part of [this discussion](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/32) – probably not _exactly_ what we want but it illustrates the size we'd expect this module to be:

```js
const { logUnhandledError } = require('@dotcom-reliability-kit/log-error');

/**
 * @typedef {object} CrashHandlerOptions
 * @property {import('process')} [process]
 *     The Node.js process to add crash handlers for.
 */

/**
 * Register a crash handler on a process.
 *
 * @public
 * @param {CrashHandlerOptions} [options = {}]
 *     Options to configure the crash handler.
 */
function registerCrashHandler(options = {}) {
	const process = options.process || global.process;
	process.on('uncaughtException', (error) => {
		logUnhandledError({ error });
		process.exit(process.exitCode || 1);
	});
}

module.exports = registerCrashHandler;
```

As outlined in [the discussion](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/32#discussioncomment-3268451), we may also want to make a few more things configurable, e.g. the exit code or allowing an app to register some of it's own teardown code:

```js
const registerCrashHandler = require('@dotcom-reliability-kit/crash-handler');

registerCrashHandler({
    exitCode: 137,
    onError: error => {
        // allow the app to inject extra behaviour when an uncaught exception occurs
    }
});
```

## Alternatives

[We documented alternatives in our design document for this work](https://financialtimes.atlassian.net/l/cp/wVjp7tL0).

## See also

CPREL-191

rowanmanning:
This is the basic functionality for an uncaught exception handler. We're going with a single option for now so that you can pass the process in (e.g. for binding to a child process).

I decided against adding an `exitCode` or `onError` option as outlined in the issue, because I think we shouldn't add things unless our teams have demostrated a need for them.

Resolves #225.

### DIFF
--- a/docs/getting-started/logging-errors.md
+++ b/docs/getting-started/logging-errors.md
@@ -176,7 +176,7 @@ There are times where an unexpected error can occur in your app and your code is
 Uncaught exception handling should never attempt to keep the app alive. If an error is thrown in a place where no handling is present then it's correct for the application to crash, _however_ if the app is crashing then we need to be sure that the fatal error is logged somewhere:
 
 > **Note**
-> This code examples is illustrative, you should not use this in your application but it helps us explain how logging fatal errors works. For a production solution, see the later examples under this heading.
+> This code example is illustrative, you should not use this in your application but it helps us explain how logging fatal errors works. For a production solution, see the later example under this heading.
 
 ```js
 process.on('uncaughtException', (error) => {
--- a/docs/getting-started/logging-errors.md
+++ b/docs/getting-started/logging-errors.md
@@ -171,8 +171,27 @@ app.get('/fruit/:name', async (request, response, next) => {
 
 ## Unhandled errors
 
-**:construction: Under construction:** this part of the documentation relates to a planned feature.<br/>
-[See the discussion here for more information on unhandled errors](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/32).
+There are times where an unexpected error can occur in your app and your code is not designed to handle that error. This is OK, and it's where the [Node.js `uncaughtException` event](https://nodejs.org/api/process.html#event-uncaughtexception) comes in.
+
+Uncaught exception handling should never attempt to keep the app alive. If an error is thrown in a place where no handling is present then it's correct for the application to crash, _however_ if the app is crashing then we need to be sure that the fatal error is logged somewhere:
+
+> **Note**
+> This code examples is illustrative, you should not use this in your application but it helps us explain how logging fatal errors works. For a production solution, see the later examples under this heading.
+
+```js
+process.on('uncaughtException', (error) => {
+    console.log(`The app is crashing because ${error.message}`);
+    process.exit(1);
+});
+```
+
+[`@dotcom-reliability-kit/crash-handler`](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/crash-handler#readme) offers a way to register a consistent error handler in your apps, which [provides as much error information as possible](#extracting-logging-information). Using this will make fatal errors in your app a lot easier to debug.
+
+It abstracts the actual uncaught exception handling code away into a neat function call:
+
+```js
+registerCrashHandler({ process });
+```
 
 
 
--- a/packages/crash-handler/README.md
+++ b/packages/crash-handler/README.md
@@ -34,7 +34,12 @@ const registerCrashHandler = require('@dotcom-reliability-kit/crash-handler');
 
 ### `registerCrashHandler`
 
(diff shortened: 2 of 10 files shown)


packages/crash-handler/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+import registerCrashHandler from '@dotcom-reliability-kit/crash-handler';
+// or
+const registerCrashHandler = require('@dotcom-reliability-kit/crash-handler');
+```
+
+### `registerCrashHandler`
+
+The `registerCrashHandler` function can be used to bind an event handler to the [Node.js `process.uncaughtException` event](https://nodejs.org/api/process.html#event-uncaughtexception). This ensures that your application logs a final message before crashing in the event on an unexpected error or promise rejection. This function should only ever be called once in your app, normally alongside all your setup code (e.g. alongside creating an Express app).
```
andygout, line 37: > normally alongside all your setup code (e.g. alongside creating an Express app)

With some of the other Reliability Kit packages, their ordering in the setup code is relevant - is that the case here (and is it worth saying so)?

packages/crash-handler/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+```js
+registerCrashHandler({
+    // Config options go here
+});
+```
+
+#### `options.process`
+
+The [Node.js Process](https://nodejs.org/api/process.html#process) object to bind the error handling event to. You may use this if you are using a child process or want to mock the process object in your tests.
```
andygout, line 60: > if you are using a child process

Does this happen much in Customer Products apps? Are there any examples you encountered as part of this work?

andygout (approved):
I have asked a couple of non-blocking questions.

packages/crash-handler/README.md
rowanmanning: There _shouldn't_ be any ordering, but I guess we're more likely to catch errors if it's one of the first things that's called. Maybe I'll add a note that there's no explicit requirement but the earlier the better.

packages/crash-handler/README.md
rowanmanning: There's some stuff around clustering where someone might want to bind to each child process from the main process (e.g. [preflight](https://github.com/Financial-Times/next-preflight/blob/main/server/cluster.js)) but tbh this also just made it much easier to test so I left it in.

docs/getting-started/logging-errors.md
```diff
@@ -171,8 +171,27 @@ app.get('/fruit/:name', async (request, response, next) => {
...
 ## Unhandled errors
 
-**:construction: Under construction:** this part of the documentation relates to a planned feature.<br/>
-[See the discussion here for more information on unhandled errors](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/32).
+There are times where an unexpected error can occur in your app and your code is not designed to handle that error. This is OK, and it's where the [Node.js `uncaughtException` event](https://nodejs.org/api/process.html#event-uncaughtexception) comes in.
+
+Uncaught exception handling should never attempt to keep the app alive. If an error is thrown in a place where no handling is present then it's correct for the application to crash, _however_ if the app is crashing then we need to be sure that the fatal error is logged somewhere:
+
+> **Note**
+> This code examples is illustrative, you should not use this in your application but it helps us explain how logging fatal errors works. For a production solution, see the later examples under this heading.
```
alexmuller, line 179: Non-blocking typo: "examples" -> "example"

alexmuller (approved):
This feels much simpler than n-express with n-raven, the docs are very clear.

docs/getting-started/logging-errors.md
rowanmanning: Good spot, fixed now

alexmuller:
Meant to post this when I reviewed but I also think "crash handler" is a really good name for this package - much more human than "uncaught exception handler".

PR #270, opened 2022-10-25
feat: add the ability to transform log data

rowanmanning:
This adds an option to specify log transforms - functions which can modify log data before it's sent. This is a foundational piece of work which is needed in order for us to provide feature parity with n-mask-logger.

A later piece of work will introduce a log transform which masks fields in exactly the same way as n-mask-logger.

### DIFF
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -12,6 +12,7 @@ A simple and fast logger based on [Pino](https://getpino.io/), with FT preferenc
     * [`Logger` configuration options](#logger-configuration-options)
       * [`options.baseLogData`](#optionsbaselogdata)
       * [`options.logLevel`](#optionsloglevel)
+      * [`options.transforms`](#optionstransforms)
       * [`options.withTimestamps`](#optionswithtimestamps)
     * [`logger.log()` and shortcut methods](#loggerlog-and-shortcut-methods)
     * [`logger.flush()`](#loggerflush)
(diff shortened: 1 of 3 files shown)


rowanmanning:
I played around with serializing all nested errors but it's quite a performance hit. See the "Errors" item in [this list](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/logger#how-different-data-is-serialized):

> Errors found in sub-properties of the log data are not serialized like this. This is for performance reasons: looping over every nested property to check if it's an error is expensive. Do not do this:
>
> ```js
> logger.info({ err: new Error('Oops') });
> // Outputs:
> // {
> //     "level": "info",
> //     "err": {}
> // }

This also matches up with [n-logger's behaviour](https://github.com/Financial-Times/n-logger#loglevel-message-meta):

> **IMPORTANT NOTE** - do not send errors as properties of other objects, e.g. `{event: 'My_EVENT', error}`. This will result in no details of the error being logged

So I think maybe to avoid further breaking changes in the migration between n-logger and this it might be best to mimic this behaviour anyway. The `transforms` options will allow someone to override default behaviour if they wish to take the slight performance hit. Shall we maybe add a feature request issue to this repo to consider trying out and properly measuring the performance of nested error serialisation?

A note on the `cause` property. If you pass in an error at the top level then the `cause` property will be serialised properly because the `serializeError` function we use _does_ recursively serialise the error cause. This is less of a perf hit because it's serialising a single property of each error rather than having to recurse over every property at each level of nesting.

andygout:
> So I think maybe to avoid further breaking changes in the migration between n-logger and this it might be best to mimic this behaviour anyway.

👍 

> Shall we maybe add a feature request issue to this repo to consider trying out and properly measuring the performance of nested error serialisation?

👍

andygout (approved):

PR #271, opened 2022-10-26
fix: loosen serialize-request type requirements

### LINKED ISSUE #269: Loosen type for serialise-request's request option
a project using Typescript that's _not_ using Express currently can't pass a request object into serialise-request (without `as unknown` type workarounds), even though the object it's passing in has the properties that are expected. the types should be loosened slightly to allow any object with the right subset of properties from `IncomingMessage`

rowanmanning:
I think this fixes #269. Our types are quite restrictive, but actually it's fine to pass in an object which looks vaguely like a request (has `headers`, `method`, and `url` properties). @apaleslimghost does this look like it'll fix your issue?

The easiest way to test might be for you to drop the generated types file into your project's `node_modules/@dotcom-reliability-kit/serialize-request/lib` folder and verify that you no longer have issues:

```ts
export = serializeRequest;
declare function serializeRequest(request: string | Request, options?: SerializeRequestOptions | undefined): SerializedRequest;
declare namespace serializeRequest {
    export { exports as default, ExpressRequest, HttpIncomingMessage, BasicRequest, Request, SerializeRequestOptions, SerializedRequestHeaders, SerializedRequestRouteParams, SerializedRequestRoute, SerializedRequest };
}
type Request = (BasicRequest & Record<string, any>) | ExpressRequest | (HttpIncomingMessage & Record<string, any>);
type SerializeRequestOptions = {
    includeHeaders?: string[] | undefined;
};
type SerializedRequest = {
    id: (string | null);
    method: string;
    url: string;
    headers: SerializedRequestHeaders;
    route?: SerializedRequestRoute | undefined;
};
declare namespace exports {
    export { exports as default, ExpressRequest, HttpIncomingMessage, BasicRequest, Request, SerializeRequestOptions, SerializedRequestHeaders, SerializedRequestRouteParams, SerializedRequestRoute, SerializedRequest };
}
type ExpressRequest = import('express').Request;
type HttpIncomingMessage = import('http').IncomingMessage;
type BasicRequest = {
    headers?: {
        [x: string]: string;
    } | undefined;
    method?: string | undefined;
    url?: string | undefined;
};
type SerializedRequestHeaders = {
    [x: string]: string;
};
type SerializedRequestRouteParams = {
    [x: string]: string;
};
type SerializedRequestRoute = {
    path: string;
    params: SerializedRequestRouteParams;
};
```

**Some testing notes for CP Reliability:** the fact that the linter passes with no issues and without changes to our own dependent modules indicates this isn't a breaking change. I also manually checked that type hinting in VS Code still works.

### DIFF
--- a/packages/serialize-request/lib/index.js
+++ b/packages/serialize-request/lib/index.js
@@ -1,5 +1,23 @@
 /**
- * @typedef {import('express').Request | import('http').IncomingMessage & {route: object, params: object}} Request
+ * @typedef {import('express').Request} ExpressRequest
+ */
+
+/**
+ * @typedef {import('http').IncomingMessage} HttpIncomingMessage
+ */
+
+/**
+ * @typedef {object} BasicRequest
+ * @property {Object<string, string>} [headers]
+ *     The request HTTP headers.
+ * @property {string} [method]
+ *     The request HTTP method.
+ * @property {string} [url]
+ *     The request URL.
+ */
+
+/**
+ * @typedef {BasicRequest & Record<string, any> | ExpressRequest | HttpIncomingMessage & Record<string, any>} Request
  */
 
 /**
@@ -35,7 +53,7 @@
  * @property {SerializedRequestHeaders} headers
  *     A subset of HTTP headers which came with the request.
  * @property {SerializedRequestRoute} [route]
- *     A subset of HTTP headers which came with the request.
+ *     The express route details.
  */
 
 /**
@@ -50,9 +68,10 @@ const DEFAULT_INCLUDED_HEADERS = ['accept', 'content-type'];
  * Serialize a request object so that it can be consistently logged or output as JSON.
  *
  * @public
- * @param {(string | Request)} request
- *     The request object to serialize. Either an Express Request object or a
- *     built-in Node.js IncomingMessage object.
+ * @param {string | Request} request
+ *     The request object to serialize. Either an Express Request object, a
+ *     built-in Node.js IncomingMessage object, or an object with the expected
+ *     `headers`, `method`, and `url` properties.
  * @param {SerializeRequestOptions} [options = {}]
  *     Options to configure the serialization.
  * @returns {SerializedRequest}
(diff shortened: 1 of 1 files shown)


apaleslimghost:
ily @rowanmanning

packages/serialize-request/lib/index.js
```diff
@@ -1,5 +1,23 @@
...
+ */
+
+/**
+ * @typedef {object} BasicRequest
+ * @property {Object<string, string>} headers
+ *     The request HTTP headers.
+ * @property {string} method
+ *     The request HTTP method.
+ * @property {string} url
+ *     The request URL.
```
apaleslimghost, line 16: **suggestion**: you're checking these properties exist already, so i think it would make sense to make them optional in this type? would help with compatibility with Apollo's request type, where they're also optional

apaleslimghost (approved):
this looks good as is, but i have one suggestion that would help with compatibility for my specific use case 😉

packages/serialize-request/lib/index.js
rowanmanning: Yeah that makes sense, technically any can be missing and we'll still try and serialize _something_. Done 👍

apaleslimghost: i see the types are optional in the copypasta in the description above but they're not in the PR diff, are you waiting for me to test it before committing?

rowanmanning: Oh they are sorry, I force pushed which confused things. In JSDoc square braces around a property name is the equivalent of `?` in TypeScript:

```js
/**
 * @typedef {object} Whatever
 * @property {string} [optionalProp]
 * @property {string} nonOptionalProp
 */
```

The new `BasicRequest` type is:

```js
/**
 * @typedef {object} BasicRequest
 * @property {Object<string, string>} [headers]
 *     The request HTTP headers.
 * @property {string} [method]
 *     The request HTTP method.
 * @property {string} [url]
 *     The request URL.
 */
```

apaleslimghost: ohhhhhh

andygout (approved):

andygout:
Thanks for thoroughly-written issue and PR description and comments to make clear what is happening with these changes.

PR #280, opened 2022-10-26
feat: add a mask transform

rowanmanning:
This should be the final piece of work needed to allow users of n-mask-logger to migrate to this library in future. It brings us to the point where we have a functional replacement for all three of our current logging libraries.

There's more information in the README and migration guide for this commit, but essentially you can now introduce the same masking-behaviour as n-mask-logger to the Reliability Kit logger:

```js
const { Logger, transforms } = require('@dotcom-reliability-kit/logger');

const logger = new Logger({
    transforms: [
        transforms.legacyMask()
    ]
});

logger.info({
    email: 'oops@ft.com'
});
// Outputs:
// {
//     "level": "info",
//     "email": "*****"
// }
```

I went with the name `legacyMask` so that we can continue to support this masking behaviour alongside a new one if we ever decide to improve it - I noticed quite a few issues while digging into the n-mask-logger source code and it's not as thorough as we might like it to be.

### DIFF
--- a/packages/logger/test/end-to-end/scripts/run-loggers-with-test-case.js
+++ b/packages/logger/test/end-to-end/scripts/run-loggers-with-test-case.js
@@ -4,8 +4,14 @@ process.env.MIGRATE_TO_HEROKU_LOG_DRAINS = 'true';
 process.env.SPLUNK_LOG_LEVEL = 'silly';
 
 const nLogger = require('@financial-times/n-logger').default;
-const reliabilityKitLogger = require('../../../lib').default;
+const MaskLogger = require('@financial-times/n-mask-logger');
+const reliabilityKitLogger = require('../../../lib');
 const testCases = require('../compatibility-test-cases');
+
+const nMaskLogger = new MaskLogger();
+const reliabilityKitMaskLogger = new reliabilityKitLogger.Logger({
+	transforms: [reliabilityKitLogger.transforms.legacyMask()]
+});
 
 // The test case ID is passed as an additional argument on the command.
 // We use this to find the test case in order to perform logging.
@@ -24,4 +30,16 @@ if (nLogger[method]) {
 // Run reliability kit logger
 if (reliabilityKitLogger[method]) {
 	reliabilityKitLogger[method](...args, { _logger: 'reliabilityKit' });
+}
+
+// Run n-mask-logger
+if (nMaskLogger[method]) {
+	nMaskLogger[method](...args, { _logger: 'nextMaskLogger' });
+}
+
+// Run reliability kit masked logger
+if (reliabilityKitMaskLogger[method]) {
+	reliabilityKitMaskLogger[method](...args, {
+		_logger: 'reliabilityKitMaskLogger'
+	});
 }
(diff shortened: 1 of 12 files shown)


packages/logger/test/end-to-end/scripts/run-loggers-with-test-case.js
```diff
@@ -25,3 +31,15 @@ if (nLogger[method]) {
 if (reliabilityKitLogger[method]) {
 	reliabilityKitLogger[method](...args, { _logger: 'reliabilityKit' });
 }
+
+// Run n-mask-logger
+if (nMaskLogger[method]) {
+	nMaskLogger[method](...args, { _logger: 'nextMaskLogger' });
+}
```
andygout, line 38: I am unclear on what exactly this file is testing. It feels a bit strange to have the existing logger packages as dependencies to Reliability Kit and also have tests dependent on them, though I am probably missing something.

Can you fill me in please?

andygout (approved):
I have added one non-blocking question.

Do we have a plan for gradually phasing out the existing logger packages, perhaps starting by marking them as deprecated and that any future requirements should involve migrating to Reliability Kit and accommodating changes there instead?

packages/logger/test/end-to-end/scripts/run-loggers-with-test-case.js
rowanmanning: Gladly! Everything in the `packages/logger/test/end-to-end` folder is designed to make us super confident that the new logger can replace n-logger and n-mask-logger.

In order to be sure of this, we need to install both of these old modules as development dependencies, then this test script will spin up a Node.js process and log the same data with both the old loggers and the new logger. It then reads the JSON logs _from `stdout`_ and compares them.

This means that we're verifying that, given the same input, n-logger vs reliability-kit/logger and n-mask-logger vs reliability-kit/logger output the same thing. You can see the actual comparisons in the `compatibility-test-cases.js` file, e.g. [here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/21c9079bdc633b95ce3ce21e735081d09068ccb0/packages/logger/test/end-to-end/compatibility-test-cases.js#L33-L44).

```js
[
	{
		id: 'n-logger-docs-1',
		description: '.log() with log level and message (n-logger docs)',

                // This defines that call that is made to each logger. This means that the logger will
                // be instantiated and have the `log` method called with `info` and `Saying hello`
		call: {
			method: 'log',
			args: ['info', 'Saying hello']
		},

                // This describes the output that is expected from each logger. We define them
                // separately because there are documented cases where they differ slightly and
                // we're OK with it.
		expectedOutput: {
			nextLogger: { level: 'info', message: 'Saying hello' },
			reliabilityKit: { level: 'info', message: 'Saying hello' }
		}
	}
]
```

If you want to chat more about it I'd be happy to go through in person 🙂 I'm quite proud of the way this works and it's made me pretty confident that a direct replacement will work for 90% of use cases.

rowanmanning: Gonna merge this as you said it's non-blocking. I'm kind of intending on writing a blog post about the testing process when we go more public with the experimental logger so will use that to explain further if necessary

PR #282, opened 2022-10-27
feat: add prettification to logs in local dev

### LINKED ISSUE #281: Prettify logs in local development
We should output prettified logs in local development. @apaleslimghost raised this and I realised I hadn't documented our intention to do this work.

Unformatted JSON logs are difficult to read which means that local development will be more difficult compared to n-logger and n-mask-logger.

## Ideal solution

We should use [pino-pretty](https://github.com/pinojs/pino-pretty) for this, it can have JSON-formatted logs piped into it which it then colourises, extracts information from, and renders in a more readable format. I think the best solution would be to integrate programmatically in the logger (see alternatives) because it'll be the best developer experience. We also want to make sure that under no circumstances can the logs become prettified in production as that'd cause them to be difficult to read in Splunk.

I'm thinking of using the following approach from the docs:

```js
const logger = pino({
    transport: {
        target: 'pino-pretty'
     }
});
```

This allows us to ensure that pino-pretty definitely isn't used in production. We can do the above based on two conditions:

1. Only do it if pino-pretty is installed. We will not install pino-pretty as a dependency of this logger but will document that teams need to install it as a development dependency in their project. We can check for the presence of pino-pretty with `require.resolve`

2. Only do it if `process.env.NODE_ENV` is `development` (we can get this from `@dotcom-reliability-kit/app-info`)

So if an app installs `pino-pretty` locally _and_ they are not in production, they will get pretty logs by default.

We should also make this configurable with an option in the logger so that a team can turn it off easily if necessary. We can also maybe allow toggling of colourisation:

```js
const logger = new Logger({
    development: {
        prettify: false, // default to `true`
        colorize: false  // default to `true`
    }
});
```

## Alternatives

1. We could not prettify logs. I don't think this is a viable alternative as engineers being able to read local logs is essential.

2. We could tell teams to pipe their application logs into the `pino-pretty` command line tool. This isn't a viable alternative with Tool Kit, as it manages the `npm start` command and does many more things than output application logs (including prefixing all application logs with tags). I don't think it's possible to do it this way.

3. We could introduce the prettifier into the Tool Kit `node` plugin. I don't think this is a good idea as it breaks modularity of Tool Kit plugins and relies on code outside of Reliability Kit to do an essential job.

4. We could install pino-pretty in the logger module, however this adds bloat to the module _and_ means that the only condition we can use is whether `process.env.NODE_ENV` is `development`. I can envisage this resulting in prettified logs being output in production a lot more easily.

rowanmanning:
This adds the ability for an app using the new logger to see prettified logs locally, which makes issues easier to debug. We configure pino-pretty if it's present to work with our logs.

This resolves #281. I decided against adding extra options to disable prettification, as teams can manually set their `NODE_ENV` to `production` if they want to inspect the raw JSON output.

The issue linked above contains more information on why I decided to solve this within the logger rather than externally or with Tool Kit.

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -1,6 +1,7 @@
 const pino = require('pino').default;
 const serializeError = require('@dotcom-reliability-kit/serialize-error');
 const { default: structuredClone } = require('@ungap/structured-clone');
+const appInfo = require('@dotcom-reliability-kit/app-info');
 
 /**
  * @typedef {"silly" | "data" | "debug" | "verbose" | "info" | "warn" | "error" | "fatal"} LogLevel
@@ -81,6 +82,34 @@ const logLevelToTransportMethodMap = {
 	verbose: { logLevel: 'debug', isDeprecated: true },
 	warn: { logLevel: 'warn', isDeprecated: false }
 };
+
+/**
+ * Whether log prettification is available. This is based
+ * on two things: the pino-pretty module being installed
+ * in the application, and the `NODE_ENV` environment
+ * variable being undefined or "development".
+ *
+ * @type {boolean}
+ */
+const PRETTIFICATION_AVAILABLE = (() => {
+	try {
+		// We have to `require` here rather than `require.resolve`
+		// which is less than ideal but otherwise this is actually
+		// impossible to test with Jest. Both technically do the
+		// same file system work though, and it's only done once
+		// when the module first loads. It's also safe to ts-ignore
+		// this one because it's never actually used directly.
+		// @ts-ignore
+		require('pino-pretty');
+
+		// If we get to this point, pino-pretty is installed because
+		// otherwise it would have errored. So we can just check for
+		// the environment being "development"
+		return appInfo.environment === 'development';
+	} catch (_) {
+		return false;
+	}
+})();
 
 /**
  * Class representing a logger.
@@ -171,6 +200,15 @@ class Logger {
 				messageKey: 'message', // This is for backwards compatibility with our existing logs
 				timestamp: withTimestamps
 			};
+			if (PRETTIFICATION_AVAILABLE) {
+				pinoOptions.transport = {
+					target: 'pino-pretty',
+					options: {
+						colorize: true,
+						messageKey: 'message'
+					}
+				};
+			}
 			this.#logTransport = pino(pinoOptions);
 			this.#logTransport.level = this.#logLevel;
 		}
--- a/package-lock.json
+++ b/package-lock.json
@@ -10127,10 +10127,10 @@
     },
     "packages/crash-handler": {
       "name": "@dotcom-reliability-kit/crash-handler",
-      "version": "0.1.1",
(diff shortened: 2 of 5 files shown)


packages/logger/lib/logger.js
```diff
@@ -82,6 +83,34 @@ const logLevelToTransportMethodMap = {
...
+const PRETTIFICATION_AVAILABLE = (() => {
+	try {
+		// We have to `require` here rather than `require.resolve`
+		// which is less than ideal but otherwise this is actually
+		// impossible to test with Jest. Both technically do the
+		// same file system work though, and it's only done once
+		// when the module first loads. It's also safe to ts-ignore
+		// this one because it's never actually used directly.
+		// @ts-ignore
+		require('pino-pretty');
```
apaleslimghost, line 103: **thought**: `pino-pretty` is the de facto standard prettifier, but there are others. hardcoding this precludes apps from choosing a different one. but that's probably fine

apaleslimghost (approved):

packages/logger/lib/logger.js
rowanmanning: Yeah I think that's a different configuration thing later. I'd rather start off being opinionated and introduce more variance as and when people need it 🙂

packages/logger/lib/logger.js
```diff
@@ -171,6 +200,15 @@ class Logger {
...
 			};
+			if (PRETTIFICATION_AVAILABLE) {
+				pinoOptions.transport = {
+					target: 'pino-pretty',
+					options: {
+						colorize: true,
+						messageKey: 'message'
+					}
+				};
+			}
```
andygout, line 211: Conditional property assignment means you could include the `transport` property in the initial assignment of `pinoOptions`, but looking at it, what you've written is much easier to follow, so ignore this comment 😄 

```js
const pinoOptions = {
	base: {},
	formatters: {
		level(label) {
			return { level: label };
		}
	},
	messageKey: 'message', // This is for backwards compatibility with our existing logs
	timestamp: withTimestamps,
	...(PRETTIFICATION_AVAILABLE && {
		transport: {
			target: 'pino-pretty',
			options: {
				colorize: true,
				messageKey: 'message'
			}
		}
	})
};
```

andygout (approved):
I have made one non-blocking comments which essentially nullifies itself 😅

andygout:
Also quite intrigued to see just how pretty these local logs look…

packages/logger/lib/logger.js
rowanmanning: Yeah call me old fashioned but I love a regular old `if` 😁

rowanmanning:
Ooh I absolutely should have included that in the PR description! OK so when running the following code:

```js

logger.info('This is a log', {
	this: 'is some data',
	nested: [{ ooh: 'pretty' }]
});
```

Here's the unprettified output:

Screenshot 2022-10-28 at 08 34 14

And here's the prettified output:

Screenshot 2022-10-28 at 08 33 42

PR #288, opened 2022-11-01
fix: correct ESM exports

rowanmanning:
This commit 8cdd4d74afe44c51504ddf0f9a3219db3ad2654a broke the exports when used in a native ESM context. I'd rather have a little more clutter in our type definitions than not be able to import modules as our users expect. Thanks @apaleslimghost for the heads up.

I'm going to investigate whether we need automated tests to ensure that our modules behave as expected in:

  * CommonJS (native)
  * ESM (native)
  * ESM via [Sucrase](https://sucrase.io/)
  * ESM via TypeScript
  * ESM via Babel???

I _really_ hate that this is the Node.js ecosystem.

### DIFF
--- a/packages/errors/lib/index.js
+++ b/packages/errors/lib/index.js
@@ -1,9 +1,7 @@
-module.exports = {
-	DataStoreError: require('./data-store-error'),
-	HttpError: require('./http-error'),
-	OperationalError: require('./operational-error'),
-	UpstreamServiceError: require('./upstream-service-error'),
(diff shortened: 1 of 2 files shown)


andygout (approved):

andygout:
> I really hate that this is the Node.js ecosystem.
➕

apaleslimghost:
there's another case to consider: ESM via Typescript/Babel/etc _but also outputting as native ESM_

PR #293, opened 2022-11-01
fix: add proper types for the default logger

rowanmanning:
This should address another issue that Kara spotted, where we're not properly defining types in this logger entrypoint.


apaleslimghost (approved):

alexmuller (approved):
I think this works in my editor but realistically I have about the same knowledge of JS types as a 6 month old.

Issue #223, opened 2022-10-11
Expand the list of default headers included in logs


rowanmanning:
I think the list of headers that we include in logs is pretty minimal and we could be getting better debug information by default rather than requiring each app to configure them.

## What problem does this feature solve?

We have slightly less information that we might need by default by only including the `accept` and `content-type` headers. We cannot expose _all_ headers and we need to consider not exposing sensitive data in our logs.

## Ideal solution

We should add some more default headers. A non-exhaustive list of suggestions:

  * `accept-encoding`
  * `accept-language`
  * `referer`
  * `user-agent`

## Alternatives

  * We could do nothing, but that means we might end up having lots of apps repeating the same config

  * We could expose the expanded header list to avoid some of the duplication, e.g. allowing users to access an `expandedHeaders` property which they can pass into the `includeHeaders` option if they want to

PR #297, opened 2022-11-02
feat: expand list of default headers included in serialized requests

andygout:
Fixes https://github.com/Financial-Times/dotcom-reliability-kit/issues/223

I checked the request headers that we see on a locally-running app (next-static) and there were none that seemed like they would be obviously vital to include by default at this stage:

```js
[
	'Host',
	'Connection',
	'sec-ch-ua',
	'sec-ch-ua-mobile',
	'sec-ch-ua-platform',
	'Upgrade-Insecure-Requests',
	'User-Agent',
	'Accept',
	'Purpose',
	'Sec-Fetch-Site',
	'Sec-Fetch-Mode',
	'Sec-Fetch-User',
	'Sec-Fetch-Dest',
	'Accept-Encoding',
	'Accept-Language',
	'Cookie'
]
```

The [HTTP headers list](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) provided by MDN is naturally very extensive, suggesting that perhaps the approach we take (as suggested by @rowanmanning) is to wait and see which headers teams using Reliability Kit add in addition to the defaults, and if there are popular choices then those can be added at a later date.

### DIFF
--- a/packages/log-error/README.md
+++ b/packages/log-error/README.md
@@ -155,8 +155,21 @@ logRecoverableError({
 
 #### `options.includeHeaders`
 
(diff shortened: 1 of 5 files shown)


rowanmanning (approved):
Looking great 🚀 I think this'll give us a lot of useful extra data by default 🙂

PR #302, opened 2022-11-03
feat: expose the default serialized headers

### LINKED ISSUE #222: Expose the default headers included in logs
The default headers we include in logs are not accessible programmatically. Maybe they should be?

In the @Financial-Times/ads Reliability Kit workshop they asked whether the default logged headers were accessible anywhere. Right now they're not which means we require any team which customises the header values to duplicate the default list. This will result in mistakes being made and less information being included by default.

## Ideal solution

We should expose the default headers so that they can be included when configuring included headers. E.g.

```js
const createErrorLogger = require('@dotcom-reliability-kit/middleware-log-errors');
const {defaultHeaders} = require('@dotcom-reliability-kit/serialize-request');

app.use(createErrorLogger({
    includeHeaders: [
        ...defaultHeaders,
        'my-custom-header'
    ]
}));

```

## Alternatives

We could not do this, but the problem will get worse as we add more default headers.

## See also

CPREL-347

rowanmanning:
This resolves #222, exposing the default headers that we use in `serializeRequest`. This should reduce the need for repetition across our apps as they can destructure this into their own `includeHeaders` config.

I decided to go for the uppercase "constant-style" naming with `DEFAULT_INCLUDED_HEADERS`. I also froze the exported object to avoid potential side-effects.

### DIFF
--- a/packages/serialize-request/lib/index.js
+++ b/packages/serialize-request/lib/index.js
@@ -59,7 +59,7 @@
 /**
  * The default request headers to include in the serialization.
  *
- * @private
+ * @public
  * @type {Array<string>}
  */
 const DEFAULT_INCLUDED_HEADERS = [
@@ -191,7 +191,14 @@ function createSerializedRequest(properties) {
 	);
 }
 
-module.exports = serializeRequest;
+exports = module.exports = serializeRequest;
+
+// We freeze this object so that we avoid any side-effects
+// introduced by the way Node.js caches modules. If this
+// array is edited within a dependent app, then any changes
+// will apply to _all_ uses of `serializeRequest`. This
+// could cause some weird issues so we lock it down.
+exports.DEFAULT_INCLUDED_HEADERS = Object.freeze([...DEFAULT_INCLUDED_HEADERS]);
 
 // @ts-ignore
 module.exports.default = module.exports;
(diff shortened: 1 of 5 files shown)


packages/serialize-request/lib/index.js
```diff
@@ -191,7 +191,14 @@ function createSerializedRequest(properties) {
...
 
-module.exports = serializeRequest;
+exports = module.exports = serializeRequest;
+
+// We freeze this object so that we avoid any side-effects
+// introduced by the way Node.js caches modules. If this
+// array is edited within a dependent app, then any changes
+// will apply to _all_ uses of `serializeRequest`. This
+// could cause some weird issues so we lock it down.
+exports.DEFAULT_INCLUDED_HEADERS = Object.freeze([...DEFAULT_INCLUDED_HEADERS]);
```
rowanmanning, line 201: question: actually would just `DEFAULT_HEADERS` make more sense? It's a little more concise and it's true in the context of `serializeRequest` 🤔

oliverturner: I think this would be fine personally

oliverturner (approved):
Lovely bit of dev-friendly ergonomics 🥰

andygout (approved):

packages/serialize-request/lib/index.js
rowanmanning: OK I've stopped second-guessing myself, going for the more explicit `DEFAULT_INCLUDED_HEADERS` and if we want to rename later then we can just alias it 😂

PR #304, opened 2022-11-07
feat: indicate that the crash handler is stable

rowanmanning:
This module is now ready for use across any Customer Products apps, after a two week trial. This PR bumps to a stable 1.0.0 release and includes the package in the central README.

Release-As: 1.0.0


andygout (approved):
Non-blocking comment; do we want to state the minimum version of n-express required to prevent Sentry being used by default or displaying a warning.

**Update:** Just revisited the migration guide which mentions this so guess it is fine for that information to live there 👍

alexmuller (approved):

PR #329, opened 2022-11-22
feat: add the ability to set a custom logger

### LINKED ISSUE #221: Make the logger configurable
In the `log-error` and `middleware-log-errors` packages, we currently use n-logger to output logs. This should maybe be a configurable option, allowing a team to pass in their own compatible logger.

In the @Financial-Times/acquisition Reliability Kit workshop, they requested that we make the logger configurable. This would allow them to use the Reliability Kit logger and logging middleware with n-mask-logger, which is a requirement for them.

## Ideal solution

I think making this a config option for our logging packages would solve this. E.g:

```js
const createErrorLogger = require('@dotcom-reliability-kit/middleware-log-errors');
const {logRecoverableError} = require('@dotcom-reliability-kit/log-error');
const MaskLogger = require('@financial-times/n-mask-logger');

app.use(createErrorLogger({
    logger: new MaskLogger(['email', 'password'])
}));

const errorWithPersonalData = new Error('Oops');
errorWithPersonalData.email = 'oops@ft.com';

logRecoverableError({
    error: errorWithPersonalData,
    logger: new MaskLogger(['email', 'password'])
});
```

## Alternatives

  * We could do nothing, but that blocks Acquisition from using Reliability Kit in certain places.

## See also

CPREL-327

rowanmanning:
Setting a custom logger is essential for teams who use log masking or their app relies on other custom logging (e.g. their logger must include certain base data).

This is more complex than I originally thought, as we can no longer trust the logging methods that we're using. Because of this I'm now wrapping all our error logging in a try/catch so that we can fall back to logging via `console.log` which is guaranteed to work. This is a similar approach to @reliability-kit/logger.

This resolves #221.

## Examples of how/why you'd use this

Log masking:

```js
const {logRecoverableError} = require('@dotcom-reliability-kit/log-error');
const MaskLogger = require('@financial-times/n-mask-logger');

logRecoverableError({
    error: new Error('Oops'),
    logger: new MaskLogger()
});
```

Adding base data:

```js
const {logRecoverableError} = require('@dotcom-reliability-kit/log-error');
const {Logger} = require('@dotcom-reliability-kit/logger');

logRecoverableError({
    error: new Error('Oops'),
    logger: new Logger({
        baseLogData: {
            example:  true
        }
    })
});
```

### DIFF
--- a/packages/log-error/README.md
+++ b/packages/log-error/README.md
@@ -1,18 +1,19 @@
 
-## @dotcom-reliability-kit/log-error
+# @dotcom-reliability-kit/log-error
 
 A method to consistently log error object with optional request information. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
 
-  * [Usage](#usage)
-    * [`logHandledError`](#loghandlederror)
-    * [`logRecoverableError`](#logrecoverableerror)
-    * [`logUnhandledError`](#logunhandlederror)
-    * [configuration options](#configuration-options)
-      * [`error`](#optionserror)
-      * [`includeHeaders`](#optionsincludeheaders)
-      * [`request`](#optionsrequest)
-  * [Contributing](#contributing)
-  * [License](#license)
+* [Usage](#usage)
+  * [`logHandledError`](#loghandlederror)
+  * [`logRecoverableError`](#logrecoverableerror)
+  * [`logUnhandledError`](#logunhandlederror)
+  * [Configuration options](#configuration-options)
+    * [`options.error`](#optionserror)
+    * [`options.includeHeaders`](#optionsincludeheaders)
+    * [`options.logger`](#optionslogger)
+    * [`options.request`](#optionsrequest)
+* [Contributing](#contributing)
+* [License](#license)
 
 
 ## Usage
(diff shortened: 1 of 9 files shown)


packages/log-error/README.md
```diff
@@ -199,6 +200,28 @@ logRecoverableError({
 > **Note**
 > There's no need to include the `x-request-id` header in this array, as this is automatically included as `request.id` in the logs.
 
+#### `options.logger`
```
andygout, line 203: Is it worth including the scenarios in this PR's description? e.g. why it might be necessary to use a custom logger (for masking or including specific base data).

andygout (approved):
Looks good - I took my time reading over to ensure I fully understamd.

I have added a couple of non-blocking questions.

packages/log-error/lib/index.js
```diff
@@ -47,7 +68,24 @@ function logError({ error, event, includeHeaders, level = 'error', request }) {
...
+		console.log(
+			JSON.stringify({
+				level: 'error',
+				event: 'LOG_METHOD_FAILURE',
+				message: `Failed to log at level '${level}'`,
+				error: serializeError(loggingError)
+			})
+		);
+		// eslint-disable-next-line no-console
+		console.log(JSON.stringify(logData));
```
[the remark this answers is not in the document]
rowanmanning: It will make it to Splunk but only if the app has Heroku log drains configured, and the only apps which have started to heavily adopt Reliability Kit are already migrated

packages/log-error/README.md
rowanmanning: Done 👍

andygout: I phrased that really badly - I meant to ask whether the examples included in the PR description (of the scenarios when it might necessary to use a custom logger) could be included in this README.

rowanmanning: Ooh! 😂 my bad. Yes probably, will do it as a separate PR

PR #337, opened 2022-12-02
docs: add a serverless (Lambda) example

alexmuller:
This example shows how Reliability Kit can be used in a Serverless framework app which is deployed to AWS Lambda.

It is not deployable in its current state because of the way the monorepo manages package-lock.json and node_modules. In order for it to be deployable, we would need to make a directory which contains a node_modules/ with every dependency (and its tree) that the Lambda requires.

---

Questions:

- I don't understand the lint/type failures, although I can reproduce them locally. They're running against `node_modules/url/url.js`. `node_modules/url/` was introduced as part of the Express example a long time ago 🤯 
- When I run `npm install` locally from the root, I do get a `node_modules/` in `examples/serverless` which is correctly gitignored. That modules directory only contains 6 packages in the `serverless` package tree. Is that expected with workspaces?

### DIFF
--- a/examples/serverless/package.json
+++ b/examples/serverless/package.json
@@ -1,0 +1,32 @@
+{
+  "name": "@dotcom-reliability-kit/example-serverless",
+  "version": "0.0.0",
+  "private": true,
+  "description": "An example serverless application which uses Reliability Kit",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Financial-Times/dotcom-reliability-kit.git",
+    "directory": "examples/serverless"
+  },
+  "homepage": "https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/examples/serverless#readme",
+  "bugs": "https://github.com/Financial-Times/dotcom-reliability-kit/issues",
+  "license": "MIT",
+  "engines": {
+    "node": "14.x || 16.x || 18.x",
+    "npm": "7.x || 8.x"
+  },
+  "scripts": {
+    "invoke": "SYSTEM_CODE='$$$-no-bizops-system-code-$$$' ./node_modules/.bin/serverless invoke local --function example --data '{\"status\": \"questionable\"}'"
+  },
+  "dependencies": {
+    "@dotcom-reliability-kit/errors": "*",
+    "@dotcom-reliability-kit/log-error": "*",
+    "@financial-times/n-serverless-logger": "0.5.1"
+  },
+  "devDependencies": {
+    "serverless": "3.25.1"
+  },
+  "volta": {
+    "node": "16.18.1"
+  }
+}
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,16 @@ updates:
       prefix-development: "chore:"
 
   - package-ecosystem: "npm"
+    directory: "/examples/serverless"
+    schedule:
+      interval: "daily"
+      time: "09:00"
+      timezone: "Europe/London"
+    commit-message:
+      prefix: "chore:"
+      prefix-development: "chore:"
+
+  - package-ecosystem: "npm"
     directory: "/resources/logos"
     schedule:
       interval: "daily"
--- a/examples/serverless/.gitignore
+++ b/examples/serverless/.gitignore
@@ -1,0 +1,1 @@
+.serverless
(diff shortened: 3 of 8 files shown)


rowanmanning (commented):
Code all looks great! I don't think the `node_modules` folder should be created but not sure why it is being. It might be that the `package-lock.json` is in an odd state due to the experiments you were doing. I'd be tempted to check out the `package-lock.json` from the `main` branch and then run `npm install` to re-add the new dependencies.

rowanmanning:
OK I take it back, the `node_modules` being created is normal, _however_ there is still a bunch of extra stuff in `package-lock.json` - running the following reduces the changes:

```
git checkout main -- package-lock.json;
npm i;
```

examples/serverless/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "serverless",
```
rowanmanning, line 2: Oh maybe this is part of the issue! This can't be named `serverless` otherwise it'll conflict with the `serverless` we're installing in `devDependencies`. I think maybe doing the following will at least make it clearer why we're getting lint errors.

First, rebase `main` onto this branch so it's up to date, then check out the `package-lock` from there:

```
git pull --rebase origin main && git checkout main -- package-lock.json
```

Then change this package name to:

```suggestion
  "name": "@dotcom-reliability-kit/example-serverless",
```

Then run `npm install` which will update the package-lock to properly link between modules so there's no conflict with the `serverless` name.

rowanmanning (changes requested):

examples/serverless/package.json
rowanmanning: Still getting url errors in the type checking though, which is odd. I'm trying to work out what's changed between the branches and I'm confused by this:

On the `main` branch, `url` was already installed via n-health:

```
npm ls --depth 999 url

dotcom-reliability-kit@0.0.0 /Users/rowan/Code/financial-times/dotcom-reliability-kit
└─┬ @dotcom-reliability-kit/example-express@0.0.0 -> ./examples/express
  └─┬ @financial-times/n-express@26.3.5
    └─┬ n-health@8.0.1
      └─┬ aws-sdk@2.1223.0
        └── url@0.10.3
```

```
npm ls --depth 999 n-health

dotcom-reliability-kit@0.0.0 /Users/rowan/Code/financial-times/dotcom-reliability-kit
└─┬ @dotcom-reliability-kit/example-express@0.0.0 -> ./examples/express
  └─┬ @financial-times/n-express@26.3.5
    └── n-health@8.0.1
```


But on the `CPREL-395-add-lambda-example` branch it is no longer picked up under `n-health`, it's only under `aws-sdk`:

```
npm ls --depth 999 url

dotcom-reliability-kit@0.0.0 /Users/rowan/Code/financial-times/dotcom-reliability-kit
└─┬ @dotcom-reliability-kit/example-serverless@0.0.0 -> ./examples/serverless
  └─┬ serverless@3.25.1
    └─┬ aws-sdk@2.1259.0
      └── url@0.10.3
```

```
npm ls --depth 999 n-health

dotcom-reliability-kit@0.0.0 /Users/rowan/Code/financial-times/dotcom-reliability-kit
└─┬ @dotcom-reliability-kit/example-express@0.0.0 -> ./examples/express
  └─┬ @financial-times/n-express@26.3.5
    └── n-health@8.0.1
```

(this is all after applying the changes suggested above)

rowanmanning: I have had this before and think I maybe solved with a `// @ts-ignore` comment or by adding `@types/<module>` to the dev dependencies, but neither will work here as we're never actually importing `url` and `@types/url` does not exist.

I don't understand why yet, but [this answer](https://stackoverflow.com/a/54277766) solved it for me – no more errors relating to the `url` package and it still reports errors in our own modules. Maybe someone with more TypeScript knowledge can help here?

alexmuller: Doh thanks! That makes sense.

alexmuller: I have absolutely no idea what's going on with `types: []` but it works for me too 😢

alexmuller: @rowanmanning I think this is passing now 😬

rowanmanning (approved):
Looks great 👍 I'm kind of ok with not understanding the fix for now, if it bites us later we'll deal with it then 😄

PR #354, opened 2022-12-21
feat: use AWS environment variables in app-info

rowanmanning:
This increases Reliability Kit's usefulness in an AWS environment, falling back to AWS and more generic variable names if the Heroku environment variables are not present. We can do this relatively easily for region, release version, and commit hash.

[Documentation on these environment variables is here](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime).

### DIFF
--- a/packages/app-info/README.md
+++ b/packages/app-info/README.md
@@ -34,11 +34,11 @@ The `appInfo` object has several properties which can be used to access applicat
 
 ### `appInfo.commitHash`
 
-Get the commit hash that the application last deployed. This will be a string (if `process.env.HEROKU_SLUG_COMMIT` or `process.env.GIT_COMMIT` is defined) or `null` otherwise.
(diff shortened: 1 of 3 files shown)


alexmuller (commented):
This is a great idea. One thing that's a little confusing is that we're using `GIT_COMMIT` here but the serverless-plugin-git-variables package seems to only reference `GIT_COMMIT_SHORT` and `GIT_COMMIT_LONG` - is that intentional?

rowanmanning:
Ah yeah I wanted to keep it more generic as that plugin allows you to set any named environment variable based on git information, but actually maybe we should go with `GIT_COMMIT_LONG` falling back to the more generic `GIT_COMMIT`?

alexmuller (approved):
Thanks!

PR #364, opened 2023-01-03
fix: use fake timers in middleware rendering snapshot test

alexmuller:
This test contains the current year for the copyright in the footer. The year changes every year so we should stub it.

### DIFF
--- a/packages/middleware-render-error-info/test/unit/lib/index.spec.js
+++ b/packages/middleware-render-error-info/test/unit/lib/index.spec.js
@@ -64,6 +64,9 @@ describe('@dotcom-reliability-kit/middleware-render-error-info', () => {
 		let response;
 
 		beforeEach(() => {
+			// The error page template contains the current year
+			jest.useFakeTimers().setSystemTime(new Date('1888-01-09'));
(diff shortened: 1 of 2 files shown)


packages/middleware-render-error-info/test/unit/lib/index.spec.js
```diff
@@ -64,6 +64,9 @@ describe('@dotcom-reliability-kit/middleware-render-error-info', () => {
 		let response;
 
 		beforeEach(() => {
+			// The error page template contains the current year
+			jest.useFakeTimers().setSystemTime(new Date('1888-01-09'));
```
andygout, line 68: `1888-01-09`

> 9 January 1888 - The Financial Times Launches

😁

andygout (approved):

packages/middleware-render-error-info/test/unit/lib/index.spec.js
alexmuller: I had no idea that's why the 1888 restaurant is called the 1888 restaurant until today 🙈

PR #404, opened 2023-02-01
Fix typo in readme

jamesnicholls:
i before e except after c


rowanmanning (changes requested):
Thanks! Well spotted 🙂 one small change: could you switch to a [conventional commit](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md#commit-type-prefixes)? Otherwise this won't get picked up by Release Please which works out the next release version and stuff. Something like:

```
docs: fix typo in readme
```

I assumed we had this set up as a linter but apparently not, I'll fix that!

rowanmanning (approved):
Thanks!

PR #411, opened 2023-02-06
chore: add a commitlint action to validate PRs

### LINKED ISSUE #405: Commit conventions are not enforced
If you're developing locally then commit conventions are enforced through a husky commit-message hook, however this doesn't catch when someone commits via the GitHub UI. I think we need to add a step to the build which fails it if any of the commits in a PR don't match our conventions, otherwise Release Please won't be able to calculate a correct release version.

**Node.js version:** All
**Impacted package(s):** All

## Steps to reproduce

1. See [this PR's](https://github.com/Financial-Times/dotcom-reliability-kit/pull/404) build passing despite it not using conventional commits

## Expected behaviour

I expect the build to fail, ideally with a useful message which points to the [relevant part of the contributing guide](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md#committing).

rowanmanning:
This ensures that PRs contain only valid commit messages. We all run this locally via Husky but if a PR is opened via edits in the GitHub interface then we rely on one of us spotting an incorrect commit message. This addresses that by failing the build.

I decided to use GitHub Actions instead of CircleCI for this for a couple of reasons:

  1. GitHub Actions has access to the range of commits that have appeared in the pull request without having to make any API calls. In CircleCI this turned out to be a lot more complex

  2. GitHub Actions are free for open source projects, which this is. So we're not consuming any FT resources to do this.

You can see an example of this build failing here: https://github.com/Financial-Times/dotcom-reliability-kit/pull/410

This resolves #405

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,21 @@
+name: Enforce commit formatting
+on: [pull_request]
+jobs:
+  run-commitlint-on-pr:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Validate all commits from the PR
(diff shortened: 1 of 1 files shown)


alexmuller (approved):
Looks good, it's a little confusing to have CircleCI and GitHub Actions doing separate things but your reasoning makes sense.

Do we want some of these status checks to be required to merge? Doesn't look like any of them are yet.

PR #437, opened 2023-02-24
docs: add support details

alexmuller:


docs/support.md
```diff
@@ -0,0 +1,30 @@
+# Reliability Kit support
+
+Reliability Kit is owned and maintained by the Customer Products Reliability team at the Financial Times.
+
+It's been designed to make our user-facing systems easier to support in production environments.
+
+However, we hope that it will be useful to other teams at the FT.
+
+## The support we provide
```
rowanmanning, line 9: I think we could add a bit somewhere which _should_ go unsaid but I think it's worth noting: our attitude to breaking changes and migrations.

> We will not make breaking changes to Reliability Kit without a corresponding major release (e.g. `v1.0.0` to `v2.0.0`). These changes will be documented in a way that helps users migrate. We will consider the scale of these migrations and try to minimise the impact on the teams using this library and, where possible, we will mark features as deprecated some time ahead of a major release.

docs/support.md
```diff
@@ -0,0 +1,30 @@
...
+
+Please contact us in [#cp-reliability on Slack](https://financialtimes.slack.com/archives/C02B89GEQHF) with
+any bug reports, feature ideas or requests for collaboration.
+
+### …to other teams at the FT
+
+Where time allows, we would love to improve Reliability Kit so that it better meets the needs of other
+users at the FT. This could include:
+
+- reviewing minor code changes that will make these packages work better for you
```
rowanmanning, line 24:
```suggestion
- reviewing minor code changes that will make these packages work better for you (please read the [contributing guide](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md) before opening pull requests)
```

rowanmanning (approved):
Amazing! Tbh I'd be happy for this to go out as-is, but let's get feedback from Alice/Jeison 😄 I have a couple of potential additions but non-blocking.

jsosa:
LGTM. Any cons for giving a timeline for PR reviews? Say we will aim to review PRs within 3 days

docs/support.md
alexmuller: Yes absolutely, thanks

rowanmanning (approved):
Really great 👏

docs/support.md
```diff
@@ -0,0 +1,48 @@
+# Reliability Kit support
+
+Reliability Kit is owned and maintained by the Customer Products Reliability team at the Financial Times.
+
+It's been designed to make our user-facing systems easier to support in production environments.
```
alicebartlett, line 5:
```suggestion
It's been designed to make other Customer Products systems easier to support in production environments.
```

docs/support.md
```diff
@@ -0,0 +1,48 @@
...
+any bug reports, feature ideas or requests for help.
+
+If you're experiencing an issue in production, message [#ft-next-support](https://financialtimes.slack.com/archives/C042NBBTM) and tag `@cp_reliability` so that we see it immediately.
+
+### …to other teams at the FT
+
+Where time allows, we would love to improve Reliability Kit so that it better meets the needs of other
+users at the FT. This could include:
+
+- reviewing minor code changes that will make these packages work better for you (please read the [contributing guide][] before opening pull requests)
```
alicebartlett, line 36: needs link

rowanmanning: It has one. It uses Markdown's [reference style links](https://www.markdownguide.org/basic-syntax/#reference-style-links), see the a couple of lines down for the actual link:

```markdown
[contributing guide]: https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md
```

Though actually maybe it is better if we just use normal links as they're more widely known.

rowanmanning: although that guide doesn't cover what we're doing here, which is using a link with an empty reference. Still valid and works but even less widely known it seems 😅

This is valid Markdown and the text will link to the reference which contains the same value:

```markdown
Hello [this] is a link. [so is this][]

[this]: https://example.com/1
[so is this]: https://example.com/2
```

rowanmanning (approved):
:shipit:

PR #440, opened 2023-03-01
feat: add filter option to the logging middleware

rowanmanning:
This allows an app to tell the error logging middleware not to log for specific errors or requests. I think this is required for us to use the error logging middleware in next-api because [it filters out certain logs based on which app has made the request](https://github.com/Financial-Times/next-api/blob/0a700435124f5c30b6b52d333acb2f1e28e0c6a8/server/utils/log-app-error.js#L8).

Without a filter for these errors we'll see a dramatic noisy increase in next-api logs

I don't want this option to start being used frequently across our apps so there's a warning in the README. I'm also sending further logs if someone misconfigures the option so that we have visibility if an app has broken logging due to use of this feature.

### DIFF
--- a/packages/middleware-log-errors/lib/index.js
+++ b/packages/middleware-log-errors/lib/index.js
@@ -65,7 +65,7 @@ function createErrorLoggingMiddleware(options = {}) {
 						return next(error);
 					}
 				} catch (/** @type {any} */ filterError) {
-					// If the filtering fails we just log the reguler error as-is but
+					// If the filtering fails we just log the regular error as-is but
 					// also log that the filtering failed
 					logRecoverableError({
 						// This is not an operational error because the filtering
--- a/packages/middleware-log-errors/lib/index.js
+++ b/packages/middleware-log-errors/lib/index.js
@@ -1,9 +1,24 @@
-const { logHandledError } = require('@dotcom-reliability-kit/log-error');
+const {
+	logHandledError,
+	logRecoverableError
+} = require('@dotcom-reliability-kit/log-error');
+
+/**
+ * @callback ErrorLoggingFilter
+ * @param {any} error
+ *     The error that will be logged.
+ * @param {import('express').Request} request
+ *     The Express request that resulted in the error.
+ * @returns {boolean | void}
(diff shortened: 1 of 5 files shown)


packages/middleware-log-errors/lib/index.js
```diff
@@ -32,10 +47,40 @@ function createErrorLoggingMiddleware(options = {}) {
...
 		try {
+			// If we have a filter then check whether the error should be logged
+			if (filter) {
+				try {
+					const shouldLog = filter(error, request);
+					if (!shouldLog) {
+						return next(error);
+					}
+				} catch (/** @type {any} */ filterError) {
+					// If the filtering fails we just log the reguler error as-is but
```
gyss, line 68:
```suggestion
					// If the filtering fails we just log the regular error as-is but
```

gyss (approved):

alexmuller (approved):
👍 Nice one. Some notes as I'm looking through this PR.

This PR is only changing the middleware-log-errors package, which is our Express middleware for error logging.

The end-to-end tests are really easy to read, thanks.

This is a [non-breaking change](https://semver.org/spec/v2.0.0.html) because it's adding an option, so the API remains compatible.

Side-note: Rowan and I have discussed in the past whether the logger forms part of the API and I think the part of Semver I have been glossing over is:

> Software using Semantic Versioning MUST declare a public API. This API could be declared in the code itself or exist strictly in documentation. However it is done, it SHOULD be precise and comprehensive.

If we don't pass in a filter option, the behaviour is the same as it was previously.

Should we be proactively alerting for `LOG_FILTER_FAILURE` across all our apps, or should we wait until we spot it?

rowanmanning:
> Should we be proactively alerting for LOG_FILTER_FAILURE across all our apps, or should we wait until we spot it?

Good question! [We already have an alert for regular log failures](https://financialtimes.splunkcloud.com/en-US/app/search/alert?s=%2FservicesNS%2Fnobody%2Fsearch%2Fsaved%2Fsearches%2FAlert%2520if%2520logging%2520fails), so will add this one to it.

CyntiBinti (approved):

PR #461, opened 2023-03-17
chore: disable exactOptionalPropertyTypes

rowanmanning:
This option causes issues when we upgrade to TypeScript 5.0. Removing it relaxes object types a little, making it valid to pass `undefined` when an object has an optional property. I think this is generally better as it allows a user to do things like

```js
registerCrashHandler({
    logger: process.env.USE_ANOTHER_LOGGER ? newLogger : undefined
});
```

### DIFF
--- a/jsconfig.json
+++ b/jsconfig.json
@@ -2,7 +2,6 @@
 	"compilerOptions": {
 		"alwaysStrict": true,
 		"checkJs": true,
-		"exactOptionalPropertyTypes": true,
 		"module": "commonjs",
 		"noEmit": true,
 		"noImplicitOverride": true,


CyntiBinti (approved):
So just to clarify: with this PR change, `undefined` would now automatically be included as one of the return value types (even if not explicitly stated), whereas before, you'd have to explicitly state `undefined` as a type, else it would throw an error?

rowanmanning:
Yep that's right, but for all types rather than just returns. So for example, let's say we have a function which has an optional string argument:

```js
/**
 * @param {string} [example]
 */
function myFunction(example) {
    // ...
}
```

Previously, despite these being semantically the same:

```
myFunction(); // Does not cause a type error
myFunction(undefined); // Causes a type error
```

After this change:

```
myFunction(); // Does not cause a type error
myFunction(undefined); // Does not cause a type error
```

PR #462, opened 2023-03-17
fix: correct a broken exports

rowanmanning:
It's invalid to reassign the `exports` global and it doesn't work as expected. I think this has only not been spotted because nobody is depending on the request serializer directly.

I only spotted this because TypeScript 5.0 can now detect this issue. So this PR blocks us being able to upgrade.


CyntiBinti (approved):
This makes sense to remove it from being re-assigned into `exports`

PR #481, opened 2023-03-31
feat: Allow for NODE_ENV that is not "production" or "development"

GeoffThorpeFT:
## Why?

- We run our pre-production systems using a `NODE_ENV` value of `test`
- The logger only activates `pino` pretty log output when the NODE_ENV is `development`

## What?

- Swapped the check for **equals**`"development` for **not-equal** `production`
- Added an extra test for `test` and adjusted the existing `development` and `production` tests

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -105,7 +105,7 @@ const PRETTIFICATION_AVAILABLE = (() => {
 		// If we get to this point, pino-pretty is installed because
 		// otherwise it would have errored. So we can just check for
 		// the environment not being "production" (which implies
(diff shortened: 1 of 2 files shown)


rowanmanning (commented):
Thanks Geoff 🙂 the build's not passing because we require conventional commits so that our release process can work out the correct version ([docs](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md#committing)). Could you change your commit message to match the format?

I think this would probably be a `feat:`-level change

GeoffThorpeFT:
Made a mess of the commit renames 😢  - will redo 😄

PR #484, opened 2023-03-31
feat: allow for NODE_ENV that is not "production" or "development"

### LINKED ISSUE #476: Can we enable more flexibility with NODE_ENV?

https://github.com/Financial-Times/dotcom-reliability-kit/blob/f4fb427b50302ae6d25b800d9688a4c6013f5752/packages/app-info/lib/index.js#L68

Only three values are handled by the logger:
1) 'production'
2) null
3) 'development'
The aim being to only show `pino` pretty logs whilst working in `development`

## Ideal solution

We make extensive use of `test` as our pre-production value for `NODE_ENV`. It would be great if that was treated the same as development 😄 so we could see the pretty logs

## Alternatives

We could change all our references from `test` to `development` but our environments and aws phases are named in the same way using `test` and `production` 🤔

GeoffThorpeFT:
## Why?

- We run our pre-production systems using a `NODE_ENV` value of `test`
- The logger only activates `pino` pretty log output when the NODE_ENV is `development`

## What?

- Swapped the check for **equals**`"development` for **not-equal** `production`
- Added an extra test for `test` and adjusted the existing `development` and `production` tests

---

Resolves #476

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -104,8 +104,9 @@ const PRETTIFICATION_AVAILABLE = (() => {
 
 		// If we get to this point, pino-pretty is installed because
 		// otherwise it would have errored. So we can just check for
-		// the environment being "development"
-		return appInfo.environment === 'development';
+		// the environment not being "production" (which implies
+		// "development", "test" or a similar pre-production term).
+		return appInfo.environment !== 'production';
 	} catch (_) {
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


rowanmanning (approved):
Looks great 👍 thanks for doing this!

kiyaGu:
@rowanmanning Geoff before he went on his holiday told me to merge this PR and test it but I don't have permission to do that. Could you please merge it and make a release? - Thanks

PR #490, opened 2023-04-04
fix: match n-logger's function serialization

### LINKED ISSUE #489: The logger does not serialize functions in the same way as n-logger
The logger errors when attempting to log a function. This doesn't match n-logger, which just excludes data that it can't serialize from the output.

**Node.js version:** all
**Impacted package(s):** logger

## Steps to reproduce

1. Import the Reliability Kit logger

    ```js
    ```

2. Try to log a function

    ```js
    logger.info({ fn: () => {} });
    ```

3. An error-level log will be output indicating that the log failed to send due to a function not being cloned

This does not match n-logger's behaviour.

## Expected behaviour

When a function is logged with n-logger, it just ignores the property and outputs nothing. This still isn't ideal but I think it's better than what we currently have as all other properties will be logged correctly.

I'd expect Reliability Kit's logger to also ignore the function property, but maybe log a warning if it's not too difficult.

rowanmanning:
This resolves #489. We were not matching n-logger in the way it serialized functions and this resulted in errors. We now match n-logger and just ignore any function properties.

I did this using the [`lossy` option of ungap's structured-clone](https://github.com/ungap/structured-clone#extra-features), which offers a few features over the native structured clone. The `lossy` option makes the clone behave more like `JSON.stringify`, ignoring properties that it can't serialize rather than throwing an error. Once Reliability Kit drops Node.js 16 support we can reassess this and perhaps move to native `structuredClone`.

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -1,6 +1,6 @@
 const pino = require('pino').default;
 const serializeError = require('@dotcom-reliability-kit/serialize-error');
-const { default: structuredClone } = require('@ungap/structured-clone');
+const { default: clone } = require('@ungap/structured-clone');
 const appInfo = require('@dotcom-reliability-kit/app-info');
 
 /**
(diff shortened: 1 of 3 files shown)


packages/logger/lib/logger.js
```diff
@@ -1,6 +1,6 @@
 const pino = require('pino').default;
 const serializeError = require('@dotcom-reliability-kit/serialize-error');
-const { default: structuredClone } = require('@ungap/structured-clone');
+const { default: clone } = require('@ungap/structured-clone');
```
rowanmanning, line 3: **note:** I renamed this to avoid confusion. In Node.js 18+ `structuredClone` is a global which doesn't exactly match the method signature of ungap's structured-clone module. I can imagine this confusing us a lot later.

PR #494, opened 2023-04-04
fix: correct manually published versions

rowanmanning:
This is a slight hack to get new versions of these packages published properly. I manually published them locally and forgot to include TypeScript definitions because I was tired and wasn't thinking.

This means that app-info 1.2.0 and logger 1.1.1 are broken and may cause build errors in apps that install them.

Adding a fix PR with an inconsequential change is the solution so that Release Please will trigger new version bumps.


alexmuller (approved):
Do you want to deprecate or withdraw those versions from npm once the new ones are available?

rowanmanning:
Great idea 👍 I'll deprecate just these versions

Issue #480, opened 2023-03-30
Serialize errors passed to the logger as object properties


rowanmanning:
This issue follows on from something @GeoffThorpeFT raised in #468 and in a discussion we had in the office.

In the documentation we explain that [errors passed into the logger will not be serialized](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/logger#log-data-serialization). This is partially for performance reasons and partially to [match n-logger's behaviour](https://github.com/Financial-Times/n-logger#loglevel-message-meta). E.g. this will not work and will log an empty object:

```js
logger.error({
    error: new Error('oops')
});
// {level: 'error', error: {}}
```

I think we should reverse this decision and serialize errors under specific circumstances.

## What problem does this feature solve?

Codebases which currently use n-logger are littered with errors being passed in as properties as above. E.g. [here](https://github.com/Financial-Times/next-eventpromo-api/blob/9573f4345ce76f9fb9eb0bed02bd13e6737b587c/server/services/eventpromo.js#L83-L88). We have no actual error information for these which makes me sad.

## Ideal solution

I think we should serialize errors passed in as properties under these specific circumstances:

  1. If the log has an `error` property and it's an instance of `Error` then serialize it
  2. If the log has an `err` property and it's an instance of `Error` then serialize it

Both of these should only be done at the very top level, so either of the following would work:

```js
logger.error({
    error: new Error('oops')
});

logger.error({
    err: new Error('oops')
});

logger.error(new Error('oops'));
```

But this would not:

```js
logger.error({
    prop: {
        error: new Error('oops')
    }
});
```

I think serializing just these two properties will catch nearly all of the places where errors are passed through as properties.

## Considerations

This issue doesn't consider the complexities of supporting an `err` error property that's serialized. We have two conflicting things which we need to think more about:

1. In the Splunk logs we want to be able to find error information easily, regardless of the system. This means that ideally we should convert err to error so that our Splunk searches don't have to consider both, however...

2. If we convert err to error will that break existing dashboards that expect err? Also what happens if somebody logs both `error` and `err` as arguments; which of these would be chosen? e.g. logger.error({ err: new Error('e1') }, new Error('e2')).

[...]

PR #499, opened 2023-04-05
feat: serialize errors passed to the logger as object properties so that error information is not lost

CyntiBinti:
PR for GitHub issue [#480](https://github.com/Financial-Times/dotcom-reliability-kit/issues/480).
[Jira ticket](https://financialtimes.atlassian.net/browse/CPREL-527)

Changes:
- Add an additional check in the logger file that checks if the parsed error item is an instance of Error AND has a sub-property of `error`, then serialize it.

Testing:
- Fixed e2e tests and all passing now
- Added a new unit test to get back to 100% coverage

To-Do:
Nil

---
Resolves: #480

### DIFF
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -515,13 +515,13 @@ Different types of data are serialized differently before being output as JSON:
     //     "level": "info",
     //     "error": {
     //         "cause": null,
-    //		   "code": "UNKNOWN",
-    //		   "data": {},
-    //		   "isOperational": false,
-    //		   "message": "Oops",
-    //		   "name": "Error",
-    //		   "relatesToSystems": [],
-    //		   "statusCode": null
+    //		"code": "UNKNOWN",
+    //		"data": {},
+    //		"isOperational": false,
+    //		"message": "Oops",
+    //		"name": "Error",
+    //		"relatesToSystems": [],
+    //		"statusCode": null
     //     }
     // }
     ```
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -489,7 +489,33 @@ Different types of data are serialized differently before being output as JSON:
       // }
       ```
 
-      Errors found in sub-properties of the log data are _not_ serialized like this. This is for performance reasons: looping over every nested property to check if it's an error is expensive. Do **not** do this:
+    Also note that errors found in sub-properties or as top-level errors are now automatically serialized as well:
+
+      ```js
+      logger.info({ error: new Error('Oops') });
+      // Outputs:
+      // {
+      //     "level": "info",
+      //     "error": {
+      //        "cause": null,
+      //		"code": "UNKNOWN",
+      //		"data": {},
+      //		"isOperational": false,
+      //		"message": "Oops",
+      //		"name": "Error",
+      //		"relatesToSystems": [],
+      //		"statusCode": null
+      //     }
+      // }
+      ```
+
+      Be mindful of passing in errors with an `err` property, as you'll have to serialize them yourself like below:
+
+      ```js
+      logger.info({ err: serializeError(new Error('Oops')) });
+      ```
+
+      ... else you'll get the following output:
 
       ```js
       logger.info({ err: new Error('Oops') });
@@ -498,12 +524,6 @@ Different types of data are serialized differently before being output as JSON:
       //     "level": "info",
       //     "err": {}
       // }
-      ```
-
-      If you _need_ to pass error objects as a property, you must serialize it yourself:
-
-      ```js
-      logger.info({ err: serializeError(new Error('Oops')) });
       ```
 
 #### Order of precedence
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -489,42 +489,42 @@ Different types of data are serialized differently before being output as JSON:
       // }
       ```
 
-    Also note that errors found in sub-properties or as top-level errors are now automatically serialized as well:
+    Errors found in sub-properties of the log data are _not_ serialized like this due to performance reasons: looping over every nested property to check if it's an error is expensive (do **not** do this). Also, be mindful of passing in errors with a custom error property e.g. `myErrorProperty`, as you'll have to serialize them yourself like below:
+
+    ```js
+    logger.info({ myErrorProperty: serializeError(new Error('Oops')) });
+    ```
+
+    ... if not, you'll get the following output:
+
+    ```js
+    logger.info({ myErrorProperty: new Error('Oops') });
+    // Outputs:
+    // {
+    //     "level": "info",
+    //     "myErrorProperty": {}
+    // }
+    ```
+    
+    The exception to this is if the sub-property name is either `error` or `err`, as these are automatically serialized. E.g.
 
       ```js
-      logger.info({ error: new Error('Oops') });
-      // Outputs:
-      // {
-      //     "level": "info",
-      //     "error": {
-      //        "cause": null,
-      //		"code": "UNKNOWN",
-      //		"data": {},
-      //		"isOperational": false,
-      //		"message": "Oops",
-      //		"name": "Error",
-      //		"relatesToSystems": [],
-      //		"statusCode": null
-      //     }
-      // }
-      ```
-
-      Be mindful of passing in errors with an `err` property, as you'll have to serialize them yourself like below:
-
-      ```js
-      logger.info({ err: serializeError(new Error('Oops')) });
-      ```
-
-      ... else you'll get the following output:
-
-      ```js
-      logger.info({ err: new Error('Oops') });
-      // Outputs:
-      // {
-      //     "level": "info",
-      //     "err": {}
-      // }
-      ```
+    logger.info({ error: new Error('Oops') });
+    // Outputs:
+    // {
+    //     "level": "info",
+    //     "error": {
+    //         "cause": null,
+    //		   "code": "UNKNOWN",
+    //		   "data": {},
+    //		   "isOperational": false,
+    //		   "message": "Oops",
+    //		   "name": "Error",
+    //		   "relatesToSystems": [],
+    //		   "statusCode": null
+    //     }
+    // }
+    ```
 
 #### Order of precedence
 
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -342,6 +342,10 @@ class Logger {
 				sanitizedLogData.error = serializeError(sanitizedLogData.error);
 			}
 
+			if (sanitizedLogData.err && sanitizedLogData.err instanceof Error) {
+				sanitizedLogData.err = serializeError(sanitizedLogData.err);
+			}
+
 			// Transform the log data
 			let transformedLogData = clone(sanitizedLogData, {
 				lossy: true
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -68,6 +68,7 @@ const appInfo = require('@dotcom-reliability-kit/app-info');
  * A map of log levels to the underlying log method that
  * should be called when a log of that level is sent, as
  * well as the deprecation status of the log level.
+ *
  * @type {Object<string, LogLevelInfo>}
  */
 const logLevelToTransportMethodMap = {
@@ -87,6 +88,7 @@ const logLevelToTransportMethodMap = {
  * on two things: the pino-pretty module being installed
  * in the application, and the `NODE_ENV` environment
  * variable being undefined or "development".
+ *
  * @type {boolean}
  */
 const PRETTIFICATION_AVAILABLE = (() => {
@@ -141,6 +143,7 @@ class Logger {
 
 	/**
 	 * Create a logger.
+	 *
 	 * @param {LoggerOptions & PrivateLoggerOptions} [options = {}]
 	 *     Options to configure the logger.
 	 */
@@ -238,6 +241,7 @@ class Logger {
 
 	/**
 	 * Create a child logger with additional base log data.
+	 *
 	 * @public
 	 * @param {LogData} baseLogData
 	 *     The base log data to add.
@@ -255,6 +259,7 @@ class Logger {
 
 	/**
 	 * Add additional log data to all subsequent log calls.
+	 *
 	 * @deprecated Please create a child logger with `createChildLogger` or use the `baseLogData` option.
 	 * @public
 	 * @param {LogData} extraLogData
@@ -275,6 +280,7 @@ class Logger {
 
 	/**
 	 * Set the `context` property for all subsequent log calls.
+	 *
 	 * @deprecated Please create a child logger with `createChildLogger` or use the `baseLogData` option.
 	 * @public
 	 * @param {LogData} contextData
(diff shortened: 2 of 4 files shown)


jsconfig.json
```diff
@@ -2,6 +2,7 @@
 	"compilerOptions": {
 		"alwaysStrict": true,
 		"checkJs": true,
+		"lib": ["esnext", "dom"],
```
CyntiBinti, line 5: This line was added I was getting compiling errors in my VS Code saying `object.hasOwn() does not belong in Object Constructor` but from stack overflowing I found this line as a fix. But let me know if this is a correct fix :)

rowanmanning: Ah I was wondering why this failed! You're ahead of the game 🙂 `Object.hasOwn()` is only supported in Node.js [16.10.0 and above](https://node.green/#ES2022-features-Object-hasOwn) so think we won't be able to use it reliably until after we drop Node.js 16 support in August 😔 there are definitely some alternatives we could try which will support Node.js 14 and earlier versions of 16

CyntiBinti: Ohhh I see 🥲  Hmm, ok can think of some others then! Thank you!

packages/logger/lib/logger.js
```diff
@@ -338,6 +338,14 @@ class Logger {
 				sanitizedLogData.message = null;
 			}
 
+			if (sanitizedLogData.error && sanitizedLogData.error instanceof Error) {
+				sanitizedLogData.error = serializeError(sanitizedLogData.error);
+			}
+
+			if (sanitizedLogData.err && sanitizedLogData.err instanceof Error) {
+				sanitizedLogData.err = serializeError(sanitizedLogData.err);
+			}
```
rowanmanning, line 347: **issue:** I think let's remove this use-case from the PR as chatted about yesterday. We should also copy this explanation onto the issue as a comment. My original issue didn't consider the complexities of supporting this. We have two conflicting things which we need to think about more:

1. In the Splunk logs we want to be able to find error information easily regardless of the system. This means that ideally we should convert `err` to `error` so that our Splunk searches don't have to consider both, _however_...

2. If we convert `err` to `error` will that break existing dashboards that expect `err`? Also what happens if somebody logs both, which would be choose? e.g. `logger.error({ err: new Error('e1') }, new Error('e2'))`.

I think the right call is not to implement anything for this case. I don't think it's [super common](https://github.com/search?q=org%3AFinancial-Times+err%3A+language%3AJavaScript&type=code&l=JavaScript) for people to use `err`  and if we add this support in a way we're not happy with then we'll be stuck with it for ages. Let's be cautious.

packages/logger/test/end-to-end/compatibility-test-cases.js
```diff
@@ -808,7 +808,16 @@ module.exports = [
 				level: 'error',
 				event: 'FAILED_TO_GET_FT_LIVE_EVENT',
 				message: null,
-				error: {} // Empty object because this is invalid, it says so in the n-logger docs
+				error: {
```
rowanmanning, line 811: **issue:** as well as this compatibility test change we'll also need to add a unit test for this new feature. The reason the build is failing is because, although all the tests are passing, the code coverage has dropped below 100%. Seen in this screen-grab:

Screenshot 2023-04-26 at 16 40 15

If you run the tests locally you can view code coverage stats in the browser which makes it easy to see what's missing. Run the following to open the report:

```sh
open coverage/lcov-report/logger/lib/logger.js.html
```

(for future reference opening `coverage/lcov-report/index.html` gives the full report)

You'd need to edit the `logger.spec.js` file, copying something like [this code block](https://github.com/Financial-Times/dotcom-reliability-kit/blob/f5622c877821ecf9919fb85643ea4549e96a74c8/packages/logger/test/unit/lib/logger.spec.js#L315-L331) and adapting to provide tests for this new conditional. I'm happy to pair on this if you want to, testing is one of my favourite programming tasks 😁 and some of the mocking we're doing is complex.

packages/logger/lib/logger.js
```diff
@@ -338,6 +338,14 @@ class Logger {
 				sanitizedLogData.message = null;
 			}
 
+			if (sanitizedLogData.error && sanitizedLogData.error instanceof Error) {
+				sanitizedLogData.error = serializeError(sanitizedLogData.error);
```
rowanmanning, line 342: **issue:** we'll need to update the documentation in the README for this too. [In this section](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/logger#log-data-serialization) we say:

> Errors found in sub-properties of the log data are _not_ serialized like this. This is for performance reasons: looping over every nested property to check if it's an error is expensive.

We'll need to update to say that we _do_ automatically serialize a top-level `error` property but no others.

packages/logger/lib/logger.js
CyntiBinti: Ah yes, I recall our convo re: this. Ok will remove these lines

packages/logger/lib/logger.js
CyntiBinti: Good point- thanks for spotting! I need to make a mental note to remember to check if the docs need to be updated when I submit a PR 😅

packages/logger/test/end-to-end/compatibility-test-cases.js
alexmuller: This looks right to me! It should just be an HTML file so you might be able to open the `coverage` directory in Finder and see what's in there.

alexmuller (commented):
This is looking really good! There's just one thing I'm seeing that I can't quite understand.

When I run this the error doesn't serialise (correct):

```
> logger.debug({foo: 'bar', test: new Error('yes')})
{"level":"debug","time":1682696094057,"foo":"bar","test":{},"message":null}
```

And when I run this it does (correctly again):

```
> logger.debug({foo: 'bar', error: new Error('yes')})
{"level":"debug","time":1682696010657,"foo":"bar","error":{"name":"Error","code":"UNKNOWN","message":"yes","isOperational":false,"relatesToSystems":[],"cause":null,"stack":"Error: yes\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:55:19)\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:32:33)\n    at deserialize (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:77:68)\n    at /Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/index.js:20:7\n    at Logger.zipLogData (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:507:10)\n    at Logger.log (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:328:36)\n    at Logger.debug (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:397:8)\n    at REPL2:1:8\n    at Script.runInThisContext (node:vm:122:12)\n    at REPLServer.defaultEval (node:repl:570:29)","statusCode":null,"data":{}},"message":null}
```

But when I run it with a property of `err` it does serialise it, but I don't see that behaviour in the code 🤔 

```
> logger.debug({foo: 'bar', err: new Error('yes')})
{"level":"debug","time":1682696040833,"foo":"bar","err":{"type":"Error","message":"yes","stack":"Error: yes\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:55:19)\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:32:33)\n    at deserialize (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:77:68)\n    at /Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/index.js:20:7\n    at Logger.log (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:338:29)\n    at Logger.debug (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:397:8)\n    at REPL5:1:8\n    at Script.runInThisContext (node:vm:122:12)\n    at REPLServer.defaultEval (node:repl:570:29)\n    at bound (node:domain:433:15)"},"message":null,"message":"yes"}
```

CyntiBinti:
> This is looking really good! There's just one thing I'm seeing that I can't quite understand.
> 
> When I run this the error doesn't serialise (correct):
> 
> ```
> > logger.debug({foo: 'bar', test: new Error('yes')})
> {"level":"debug","time":1682696094057,"foo":"bar","test":{},"message":null}
> ```
> 
> And when I run this it does (correctly again):
> 
> ```
> > logger.debug({foo: 'bar', error: new Error('yes')})
> {"level":"debug","time":1682696010657,"foo":"bar","error":{"name":"Error","code":"UNKNOWN","message":"yes","isOperational":false,"relatesToSystems":[],"cause":null,"stack":"Error: yes\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:55:19)\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:32:33)\n    at deserialize (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:77:68)\n    at /Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/index.js:20:7\n    at Logger.zipLogData (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:507:10)\n    at Logger.log (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:328:36)\n    at Logger.debug (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:397:8)\n    at REPL2:1:8\n    at Script.runInThisContext (node:vm:122:12)\n    at REPLServer.defaultEval (node:repl:570:29)","statusCode":null,"data":{}},"message":null}
> ```
> 
> But when I run it with a property of `err` it does serialise it, but I don't see that behaviour in the code 🤔
> 
> ```
> > logger.debug({foo: 'bar', err: new Error('yes')})
> {"level":"debug","time":1682696040833,"foo":"bar","err":{"type":"Error","message":"yes","stack":"Error: yes\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:55:19)\n    at unpair (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:32:33)\n    at deserialize (/Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/deserialize.js:77:68)\n    at /Users/alex.muller/ft/dotcom-reliability-kit/node_modules/@ungap/structured-clone/cjs/index.js:20:7\n    at Logger.log (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:338:29)\n    at Logger.debug (/Users/alex.muller/ft/dotcom-reliability-kit/packages/logger/lib/logger.js:397:8)\n    at REPL5:1:8\n    at Script.runInThisContext (node:vm:122:12)\n    at REPLServer.defaultEval (node:repl:570:29)\n    at bound (node:domain:433:15)"},"message":null,"message":"yes"}
> ```

Oh that's quite interesting/odd behaviour 🤔  Will re-run this `err` scenario you've raised and try and see why it's serializing it's this particular sub-property as it shouldn't as far as I'm aware of - thanks!

alexmuller:
Even more weirdly I added a `console.log("SERIALISING ERROR PROPERTY");` inside your new conditional and this log only outputs for the `error` property - which is what we expect. So perhaps something else is serialising the `err` 🤯. We can look at this next week!

packages/logger/README.md
```diff
@@ -489,23 +489,43 @@ Different types of data are serialized differently before being output as JSON:
       // }
       ```
 
-      Errors found in sub-properties of the log data are _not_ serialized like this. This is for performance reasons: looping over every nested property to check if it's an error is expensive. Do **not** do this:
+    Also note that errors found in sub-properties or as top-level errors are now automatically serialized as well:
```
rowanmanning, line 492: **suggestion:** reading only this line it suggests that we now serialize _all_ top-level properties which isn't true. I'd be tempted to leave this as-is but note the one exception. Maybe something like:

```suggestion
    Errors found in sub-properties of the log data are _not_ serialized like this except when the property name is `error`. This is for performance reasons: looping over every nested property to check if it's an error is expensive. Do **not** do this:
```

We'd then want to change the example below back, but because we have inconsistencies with `err` we probably want to change to something we definitely don't serialize, e.g.

> ```js
>  logger.info({ myErrorProperty: new Error('Oops') });
>  // Outputs:
>  // {
>  //     "level": "info",
>  //     "myErrorProperty": {}
>  // }

Same goes for the little bit after this, let's make it super clear by removing the `err` property from the examples and change it to something that we'll never serialize (e.g. `myErrorProperty`)

rowanmanning (commented):
Annoyingly we might need to reconsider the `err` property as serializing it differently is [built-in Pino behaviour](https://getpino.io/#/docs/api?id=errorkey-string). Maybe this means we should just serialize `err` in place and not change the property name. There's a thread in the team Slack to chat about it, would be good to get both of your thoughts @alexmuller and @CyntiBinti.

I also spotted a few other issues with the PR generally as comments below 🙂

packages/logger/README.md
```diff
@@ -489,22 +489,42 @@ Different types of data are serialized differently before being output as JSON:
...
 
       ```js
-      logger.info({ err: serializeError(new Error('Oops')) });
-      ```
+    logger.info({ error: new Error('Oops') });
+    // Outputs:
+    // {
+    //     "level": "info",
+    //     "error": {
+    //         "cause": null,
```
CyntiBinti, line 517: This line is perfectly aligned on my local IDE, but when I push my changes to GitHub it's then mis-aligned 😭

packages/logger/README.md
CyntiBinti: Good point, will change to clarify the sub-property serialization

CyntiBinti:
100% test coverage with new unit test for `err` sub-property

image

rowanmanning (approved):
This is looking great! Thanks for persevering through so much switching back and forth on the `error` / `err` thing – I probably could have refined the ticket a bit more ahead of time to avoid this but you live and learn 😄

Happy for you to merge after a final rebase! Thanks for getting this work done

packages/logger/lib/logger.js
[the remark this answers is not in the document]
CyntiBinti: Thank you! :)

PR #511, opened 2023-04-13
feat: created a new npm package, updated the README

CyntiBinti:
Created a new prototype ESLint tool for the reliability kit so that we can better catch bugs before they reach production. Once this PR is merged, it will be run against the 5 following repo's (chosen as they're pretty active and representative of Customer Products):
- [next-product](https://github.com/Financial-Times/next-product)
- [next-profile](https://github.com/Financial-Times/next-profile)
- [next-ads-api](https://github.com/Financial-Times/next-ads-api)
- [next-comments-api](https://github.com/Financial-Times/next-comments-api)
- [next-appetite](https://github.com/Financial-Times/next-appetite)

Relates to:
- [Jira](https://financialtimes.atlassian.net/browse/CPREL-538)

Changes:
- Created a new npm package
- Updated the README file
- Copied over and reviewed the current [next-eslint-config](https://github.com/Financial-Times/eslint-config-next) `index.js` file (tweaked to our needs instead)

Testing:
- I've added a comment block with my intentions for testing in the `index.spec.js` file

To-Do:
- Add all the agreed ESLint rules from the [audit sheet](https://docs.google.com/spreadsheets/d/13ylBX9lHapx9cz-1KO7K0kDXXXVtaF9TsOlxNjxspn0/edit#gid=0)
- Write unit tests to ensure this package won't throw an error if ran in another repo


packages/eslint-tool/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# See https://help.github.com/articles/about-codeowners/ for more information about this file.
+
+- @financial-times/reliability
```
CyntiBinti, line 3: Don't know if this file is necessary? or if I've used the right name path for our team?

packages/eslint-tool/lib/index.js
```diff
@@ -0,0 +1,81 @@
...
+		es6: true,
+		mocha: true,
+		node: true
+	},
+	parserOptions: {
+		ecmaVersion: 'latest',
+		sourceType: 'module'
+	},
+	rules: {
+		RuleToAdd: 'RuleToAdd'
```
CyntiBinti, line 15: Will add the agreed rules here in the next commit

packages/eslint-tool/lib/index.js
```diff
@@ -0,0 +1,81 @@
+const path = require('path');
+
+const config = {
+	env: {
+		browser: true,
+		es6: true,
```
CyntiBinti, line 6: Was reading through the eslint docs and didn't know if this should be bumped up to the latest [es2022](https://eslint.org/docs/latest/use/configure/language-options) or leave as es6?

packages/eslint-tool/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@dotcom-reliability-kit/eslint-tool",
+  "version": "0.0.0",
+  "description": "TODO",
```
CyntiBinti, line 4: Just noticed this. Will update in next commit

packages/eslint-tool/README.md
```diff
@@ -0,0 +1,56 @@
...
+A prototype linting tool, specifically focussed on enhancing code quality and proactively catching errors/bugs before they make it into production. This is not a replacement for the [@financial-times/eslint-config-next](https://github.com/Financial-Times/eslint-config-next), but it can be installed alongside it or as an extension in it's config. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
+
+- [Usage](#usage)
+  - [Static Code Analysis](#static-code-analysis)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Usage
+
+---
```
rowanmanning, line 12: **suggestion:** these [horizontal rules](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hr) after headings aren't necessary as they result in some invalid sectioning in the output HTML. They should used to denote a thematic break between paragraphs rather than to underline titles as screen readers will pick them up.

packages/eslint-tool/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Install @dotcom-reliability-kit/eslint-tool as a dependency:
+
+```
+npm install --save @dotcom-reliability-kit/eslint-tool
+```
```
rowanmanning, line 18: **suggestion:** we should make sure we say to install this as a development dependency, we don't want it to be installed in production. So:

> Install @dotcom-reliability-kit/eslint-tool as a development dependency:
>
> ```
> npm install --save-dev @dotcom-reliability-kit/eslint-tool
> ```

packages/eslint-tool/README.md
```diff
@@ -0,0 +1,56 @@
...
+module.exports = {
+extends: ['@dotcom-reliability-kit/eslint-tool']
+}
+```
+
+Avoid using .eslintrc file format (without a file extension), as [this has been depracated](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats).
+
+You can find the docs for ESLint shareable configs [here](https://eslint.org/docs/latest/extend/shareable-configs).
+
+### Static Code Analysis
```
rowanmanning, line 32: **question:** I'm unsure how much we should document use of ESLint, because most of our teams use it via [Tool Kit's eslint plugin](https://github.com/Financial-Times/dotcom-tool-kit/tree/main/plugins/eslint). Thoughts on this? Maybe we just link out to documentation? Unsure

packages/eslint-tool/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+See the [central contributing guide for Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md).
+
+## License
+
+---
+
+Licensed under the [MIT](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/LICENSE) license.
+
+Copyright © 2022, The Financial Times Ltd.
```
rowanmanning, line 56: **suggestion:** I'm always unsure whether we should be updating old ones too, but let's add the current year for this 😄

```suggestion
Copyright © 2023, The Financial Times Ltd.
```

packages/eslint-tool/README.md
```diff
@@ -0,0 +1,56 @@
+## @dotcom-reliability-kit/eslint-tool
+
+A prototype linting tool, specifically focussed on enhancing code quality and proactively catching errors/bugs before they make it into production. This is not a replacement for the [@financial-times/eslint-config-next](https://github.com/Financial-Times/eslint-config-next), but it can be installed alongside it or as an extension in it's config. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
+
+- [Usage](#usage)
+  - [Static Code Analysis](#static-code-analysis)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Usage
```
rowanmanning, line 10: **praise:** great usage guide, very clear and like the linking out to official docs rather than including too much detail 👏

rowanmanning (commented):
Thanks Cynthia! This is great 🙂 I've got a few comments, I've used [conventional comments](https://conventionalcomments.org/) to make it clear when something's a suggestion vs just me nitpicking so you know if there's any action required!

Let me know if you have any questions on any of these points and I'm always happy to be disagreed with if you have different thoughts.

packages/eslint-tool/lib/index.js
rowanmanning: I'm unsure, maybe we don't get to make this decision because each project should get to decide? But really don't know. Maybe requires more research to see how it works when you apply multiple configs on top of each other

CyntiBinti: Ok will look further into this :)

.release-please-manifest.json
```diff
@@ -1,11 +1,12 @@
 {
-  "packages/app-info": "1.2.1",
-  "packages/crash-handler": "1.1.4",
-  "packages/errors": "1.3.0",
-  "packages/log-error": "1.5.4",
-  "packages/logger": "1.1.2",
-  "packages/middleware-log-errors": "1.5.4",
-  "packages/middleware-render-error-info": "1.1.17",
-  "packages/serialize-error": "1.1.4",
-  "packages/serialize-request": "1.1.1"
```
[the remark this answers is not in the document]
CyntiBinti: Package to add back in: "packages/eslint-config": "0.0.0"

packages/eslint-tool/README.md
CyntiBinti: Ohh thanks - this is helpful to know #TIL

packages/eslint-tool/README.md
CyntiBinti: Ah ok - yeah makes sense to not link various docs and better to have one reference that people go to that's relevant for our org. I'll remove this line to avoid confusion

.release-please-manifest.json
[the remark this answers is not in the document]
CyntiBinti: Somehow my fix has resulted in more white spaces 😭  but my file on VS Code doesn't look affected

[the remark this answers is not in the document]
rowanmanning: Ah this is one of those rare cases where we use two-spaces for indentation 🙂 it's because Release Please automatically updates this file and resets the indentation to two-spaces each time, it's not what we'd generally use but Release Please doesn't give us the option

packages/eslint-config/README.md
```diff
@@ -0,0 +1,46 @@
...
+  - [Static Code Analysis](#static-code-analysis)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Usage
+
+Install @dotcom-reliability-kit/eslint-config as a development dependency:
+
+```
+npm install --save-dev @dotcom-reliability-kit/eslint-tool
```
rowanmanning, line 15: **issue:** just one missed instance of `eslint-tool` here

.release-please-manifest.json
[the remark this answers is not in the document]
CyntiBinti: Ok I'll try and delete the extra space and see if that fixes it - thanks! 👍🏽

rowanmanning (commented):
Looking great, thanks for making the changes 🙂 there's just one remaining reference to `eslint-tool` which was missed

.release-please-manifest.json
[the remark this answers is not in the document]
CyntiBinti: Fixed by manually deleting down the indents from 4 to 2! Thank you 😁

packages/eslint-config/README.md
CyntiBinti: Oh pesky one slipped through. Will do a final search through. Thanks!

packages/eslint-config/lib/index.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+if (packageJson && packageJson.eslintConfig) {
+	Object.assign(config.env, packageJson.eslintConfig.env);
+	Object.assign(config.parserOptions, packageJson.eslintConfig.parserOptions);
+	Object.assign(config.rules, packageJson.eslintConfig.rules);
+	Object.assign(config.globals, packageJson.eslintConfig.globals);
+	Object.assign(config.settings, packageJson.eslintConfig.settings);
+}
+
```
rowanmanning, line 80: **suggestion:** I don't think we actually need any of this, as this stuff is handled by eslint-config-next. We can probably simplify a lot by deferring React/Preact/Jest decisions to them.

packages/eslint-config/lib/index.js
```diff
@@ -0,0 +1,81 @@
...
+		RuleToAdd: 'RuleToAdd'
+	},
+	globals: {
+		cy: true,
+		cypress: true,
+		Cypress: true,
+		fetch: true,
+		requireText: true
+	},
+	plugins: ['no-only-tests'],
```
rowanmanning, line 24: **issue:** this plugin isn't installed as a dependency so it won't work, but I can see it's already included by eslint-config-next so it's not necessary for us to include it

```suggestion
```

packages/eslint-config/lib/index.js
```diff
@@ -0,0 +1,81 @@
+const path = require('path');
+
+const config = {
```
rowanmanning, line 3: Actually there's a lot in here we probably don't need, like the `globals`, `extends`, and `overrides` properties. All of these will be inherited from the eslint-config-next package. We can probably get away with only actually setting `env` and `rules` 🙂

rowanmanning (commented):
Sorry I think I completely missed the actual library code in my first review, but I think we can strip out a lot of stuff here because of the way ESLint inheritance works. A lot of the complexity in here will be handled by eslint-config-next 🙂

packages/eslint-tool/lib/index.js
CyntiBinti: Decided to bump up to es2022 as (from what I've read) this means that this environment will provide predefined global variables from ECMAScript 2022 globals.

Which I interpret as 🤔 : any new global variables that have been released since es6 (ie pre-2016) up to es2022 would also be included in the ESLint check. But correct me if I've interpreted this incorrectly 🙇🏽‍♀️

packages/eslint-config/lib/index.js
CyntiBinti: So I umm'd and ahh'd on this too and thought of a scenario where an app only installs our reliability-kit-eslint-config and didn't already have the next-eslint-config. Would this block of code need to be present then or not?

rowanmanning: I think we shouldn't duplicate the code across configs, and I don't think we've looked into React / JSX usage enough to make decisions about how we lint it yet, whereas the core rules we've gone thorough and made decisions about. If I'm honest I don't like this code anyway and think we should remove from next-eslint-config at a later date 😁 maybe having a separate config for React-based apps.

packages/eslint-tool/test/unit/lib/index.spec.js
```diff
@@ -0,0 +1,16 @@
...
+	});
+});
+
+/*
+Tests to Write:
+
+- Read the index.js file to test that is has valid code that can be extended in another apps eslintrc file
+- This can be tested by:
+	1 - something is being exported
+	2 - that this exported 'something' is an object data structure
```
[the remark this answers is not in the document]
CyntiBinti: Ohh thank for sharing these links. Didn't know you could do this - very useful 😸

.release-please-manifest.json
```diff
@@ -1,11 +1,12 @@
...
+  "packages/app-info": "1.2.1",
+  "packages/crash-handler": "1.1.4",
+  "packages/errors": "1.3.0",
+  "packages/log-error": "1.5.4",
+  "packages/logger": "1.1.2",
+  "packages/middleware-log-errors": "1.5.4",
+  "packages/middleware-render-error-info": "1.1.17",
+  "packages/serialize-error": "1.1.4",
+  "packages/serialize-request": "1.1.1",
+  "packages/eslint-config": "0.0.0"
```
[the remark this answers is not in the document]
CyntiBinti: Updated now and all back to version 2.0.0

rowanmanning (approved):
Looking great to me! Congrats on your first Reliability Kit package 🥳 hopefully the first of many

CyntiBinti:
> Looking great to me! Congrats on your first Reliability Kit package 🥳 hopefully the first of many

Yay!! 💃🏽 TYSM!

PR #524, opened 2023-04-19
docs: added eslint-config to usage package summary list

CyntiBinti:
Small PR as forgot to update the root README with the new eslint package in the usage's summary list


rowanmanning (commented):
We don't generally add packages to this list unless they're stable 🙂 so we can park this until we're ready to go to v1.0.0

CyntiBinti:
> We don't generally add packages to this list unless they're stable 🙂 so we can park this until we're ready to go to v1.0.0

Ah ok cool. This sounds good to me 😌 Will delete this PR for now, and can always update this README if the prototype gets released

PR #525, opened 2023-04-19
fix: add eslint as a peer dependency

rowanmanning:
This adds eslint as a peer dependency, because otherwise we get some incompatibilities when installing within a repo. I just tested in dotcom-debug-user and got some cryptic errors because it wasn't using the latest version of ESLint.

I didn't consider this when reviewing the original PR :)

By adding a peerDependency of `eslint@8.27.0` we ensure that all the rules we specify actually exist within the app's installed version of ESLint. The one that immediately threw an error in dotcom-debug-user was https://eslint.org/docs/latest/rules/no-new-native-nonconstructor

Peer dependencies are designed for this case. They're meant to be used when a module _requires_ that it's installed alongside a specific version of another module. There's a good blog post on the subject here if not familiar: https://nodejs.org/en/blog/npm/peer-dependencies

### DIFF
--- a/packages/eslint-config/package.json
+++ b/packages/eslint-config/package.json
@@ -14,5 +14,8 @@
     "node": "16.x || 18.x",
     "npm": "7.x || 8.x"
   },
-  "main": "lib"
+  "main": "lib",
+  "peerDependencies": {
+    "eslint": ">=8.27.0"
+  }
 }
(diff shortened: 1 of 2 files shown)


packages/eslint-config/package.json
```diff
@@ -14,5 +14,8 @@
     "node": "16.x || 18.x",
     "npm": "7.x || 8.x"
   },
-  "main": "lib"
+  "main": "lib",
+  "peerDependencies": {
+    "eslint": "^8.27.0"
```
rowanmanning, line 19: The reason I chose `8.27.0` rather than the latest version is that this is the minimum version that we need installed. This is because [this rule](https://eslint.org/docs/latest/rules/no-new-native-nonconstructor) was only added in `8.27.0` so any earlier version of `eslint` will error

CyntiBinti (approved):

packages/eslint-config/package.json
CyntiBinti: Fair point 😌

Issue #516, opened 2023-04-17
Conflicts with lambda-logger but not with n-logger


GeoffThorpeFT:
**Node.js version:** v14
**Impacted package(s):** logger

## Steps to reproduce
run this makefile script:
```
testLoggers:
	echo =========================== no loggers ==================================
	npm uninstall @dotcom-reliability-kit/logger
	npm uninstall @financial-times/n-logger
	npm uninstall @financial-times/lambda-logger
	node test-loggers false false false

	echo =========================== lambda-logger ONLY ==================================
	npm uninstall @dotcom-reliability-kit/logger
	npm uninstall @financial-times/n-logger
	npm install @financial-times/lambda-logger
	node test-loggers false false true

	echo =========================== n-logger ONLY ==================================
	npm uninstall @dotcom-reliability-kit/logger
	npm install @financial-times/n-logger
	npm uninstall @financial-times/lambda-logger
	node test-loggers false true false

	echo =========================== n-logger PLUS lambda-logger ==================================
	npm uninstall @dotcom-reliability-kit/logger
	npm install @financial-times/n-logger
	npm install @financial-times/lambda-logger
	node test-loggers false true true

	echo =========================== kit logger ONLY==================================
	npm install @dotcom-reliability-kit/logger
	npm uninstall @financial-times/n-logger
	npm uninstall @financial-times/lambda-logger
	node test-loggers true false false

	echo =========================== kit logger PLUS lambda logger ==================================
	npm install @dotcom-reliability-kit/logger
	npm uninstall @financial-times/n-logger
	npm install @financial-times/lambda-logger
	node test-loggers true false true

	echo =========================== kit logger PLUS n-logger ==================================
	npm install @dotcom-reliability-kit/logger
	npm install @financial-times/n-logger
	npm uninstall @financial-times/lambda-logger
	node test-loggers true true false

	echo =========================== ALL THREE loggers ==================================
	npm install @dotcom-reliability-kit/logger
	npm install @financial-times/n-logger
	npm install @financial-times/lambda-logger
	node test-loggers true true true
```

This is the code for `test-loggers`:
```
const testLoggers = () => {
	const [, , useKitLogger, useNLogger, useLambdaLogger] = process.argv;
	console.log("@@@@@@@ start @@@@@@@@@")
	console.log(useKitLogger, useNLogger, useLambdaLogger);
	if (useKitLogger === 'true') {
[...]

Issue #474, opened 2023-03-27
Provide ability to config pino-pretty


j-dav:
## What problem does this feature solve?

[Slack thread](https://financialtimes.slack.com/archives/C02B89GEQHF/p1679911864017759)

- Currently the logger package doesn't allow passing of custom pino-pretty CLI commands i.e colorize, singleLine. It can get rather frustrating to read on multi line and grepping is made difficult by the multiline printout. Being able to grep and search for certain things in the terminal is quite important for finding relevant logs in development process.
- It will solve the above point by allowing users to configure their pino-pretty with [opions](https://www.npmjs.com/package/pino-pretty#cliargs:~:text=%60%0A%7D)-,Options,-The%20options%20accepted) outlined by pino-pretty docs.

A clear and concise description of what the problem is. e.g. "I'm always frustrated when [...]"

## Ideal solution
- Provide ability to pass in a `pino-prettyrc` configuration file and/or ability to add it to the command line args like `"nodemon web-server.js ./node_modules/.bin/pino-pretty --singleLine"`

Requires change to the logger package

rowanmanning:
Hey Geoff :) nice work on the migration so far! To make it clear, none of the solutions I proposed will require further code changes if you don't need the `singleLine` logger configuration, and we won't be releasing a breaking change any time soon. It's more an opt-in feature.

The environment variable options would allow the default logger to pick up the values 👍 so options 3 and 4 would not require you to instantiate a new logger if you wanted to set the `singleLine` pino-pretty option

j-dav:
> The environment variable options would allow the default logger to pick up the values 👍 so options 3 and 4 would not require you to instantiate a new logger if you wanted to set the `singleLine` pino-pretty option

Thanks for looking into this Rowan, I like ENV implementation. Being able to add CLI args via option 4 by disabling and just using your local globally installed pino-pretty package seems like the most straightforward option and gives the user freedom to do what suits.

rowanmanning:
@j-dav the prettifier is now able to be disabled as discussed. [See the documentation here on how to do this](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/logger#optionswithprettifier). This should allow you to configure pino-pretty however you like by disabling it and directly using the CLI.

PR #541, opened 2023-04-27
fix: check for supported pino-pretty versions

rowanmanning:
I'm gonna start by saying that I hate this. This is a first attempt at fixing issue #516 which is caused by the app installing a version of pino-pretty older than v10. This can easily happen unfortunately.

What I hate about this solution is how complex it is, but I'm not sure if there's a nice alternative. What this solution does is loads in the pino-pretty `package.json` file, checks the version, and ensures that it meets our minimum version requirements. If this doesn't work then we log a warning and don't try to use the prettifier. It's just so complex!

Some alternatives I've considered are:

1. **Get rid of the pino-pretty integration altogether** and make log prettification a problem for the app which uses the Reliability Kit logger. This has the added benefit of solving #474. It introduces a new issue that any apps which prettify logs will have to add configuration to expect a `message` rather than `msg` (pino's gross default) property. We'd also have to release a major version to drop the prettifier.

2. **Define pino-pretty v10 as a `peerDependency`** with an `peerDependencyMeta` of `optional: true`. This has the benefit of not installing pino-pretty alongside the Reliability Kit logger but there is a warning output when you install an incompatible version of pino-pretty. The issue is that people mostly ignore npm install warnings so I think we'll still get support requests.

Thoughts? I think alternative 2 might be the most sensible approach with alternative 1 as a further future change? Ugh.

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -83,6 +83,25 @@ const logLevelToTransportMethodMap = {
 	warn: { logLevel: 'warn', isDeprecated: false }
 };
 
+// IMPORTANT: increasing this hard-coded value is a breaking
+// change, because apps which currently use a pino-pretty
+// version lower than the new value will have to upgrade
+const PINO_PRETTY_MIN_VERSION = 10;
+
+/**
+ * Send a backup log using `console.log`, for when all else fails.
+ *
+ * @param {Object<string, any>} logData
+ *     The information to log.
+ */
+function sendBackupLog(logData) {
+	// We allow use of `console.log` here to ensure that critical
+	// logging failures are caught and logged. This ensures that we
+	// know if an app has broken logging.
+	// eslint-disable-next-line no-console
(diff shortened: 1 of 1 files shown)


wheresrhys:
I quite like alternative 2. I think a big part of the problem for us is that there weren't many clues as to what the issue was so it took some painful investigation - an install warning might provide just enough of a signpost to be helpful. @GeoffThorpeFT what do you think?

Alternative 2 could also be made `optional: false` in a future major release (could it...?)

kavanagh:
I prefer optional 2.

FWIW your implementation isn't that far off the documented approach for `optionalDependencies`

GeoffThorpeFT:
In light of the problems I had with the incompatibilities I am happy with version 2 because I would have (eventually 😄 ) noticed the warnings

rowanmanning:
Excellent thanks all for contributing! I'll go with option #2 as a separate PR and close this one 👍

PR #542, opened 2023-04-28
feat: add support for npm 9

jkerr321:
As part of the node.js 18 migration apps will also be migrating to npm 9

Reliability-kit already supports node.js 18 so this commit adds support for npm 9 to avoid `npm WARN EBADENGINE Unsupported engine` alerts when npm 9 using apps install reliability-kit packages


CyntiBinti (approved):
Thank you!! 🤗

PR #551, opened 2023-05-02
feat: allow disabling log prettification

### LINKED ISSUE #474: Provide ability to config pino-pretty

rowanmanning:
This resolves #474 by allowing an application to completely disable the internal log prettification. We now allow setting a `withPrettifier` option to `false`. This is also possible with a new environment variable named `LOG_DISABLE_PRETTIFIER`.

This is also required before I can properly work on #516.

### DIFF
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -10,6 +10,7 @@ A simple and fast logger based on [Pino](https://getpino.io/), with FT preferenc
       * [`options.baseLogData`](#optionsbaselogdata)
       * [`options.logLevel`](#optionsloglevel)
       * [`options.transforms`](#optionstransforms)
+      * [`options.withPrettifier`](#optionswithprettifier)
       * [`options.withTimestamps`](#optionswithtimestamps)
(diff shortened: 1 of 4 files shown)


CyntiBinti (approved):
LGTM! :)

PR #564, opened 2023-05-05
Create a new script that will run eslint against multiple repo's

CyntiBinti:
As part of wider OKR Q2 work for [Jira ticket](https://financialtimes.atlassian.net/jira/software/c/projects/CPREL/boards/1426?modal=detail&selectedIssue=CPREL-539).

Changes:
Used the [Eslint Node.js API](https://eslint.org/docs/latest/integrate/nodejs-api#-eslintlintfilespatterns) to help create this script as they have a pre-configured ESLint class, where you can make an instance of it and configure it to your needs.

Testing:
- Have ran the script and compared to data collected manually and they match with the number of errors caught (manual sample of 5 bronze repo's)

Considerations to Bear in Mind:
- The script version also catches `error Parsing error: Unexpected token` type errors, so these will need to be filtered out of the final audit results
- We think the ESLint class does some extra optimisation under the hood, as it will throw a memory heap error if linting too many repos (e.g. 500+). BUT, if the below command is used via the terminal, then it lints just fine:

> `NODE_OPTIONS=--max_old_space_size=8000 eslint --no-eslintrc --config /Users/cynthia.mbulu/Documents/code/dotcom-reliability-kit/packages/eslint-config/lib/index.js ./repos-old`

Future Features:
- Future work to enhance this script would be to parse the linted output into a file (e.g. *.json, *.csv, *.txt), and then for that to be further manipulated, filtered, and sorted ie by number of errors in ascending order (something like this [exec function](https://github.com/Financial-Times/n-heroku-tools/blob/7614b5aaa17119a3a32ea7e5b9135de0257ebd93/scripts/generate-docs.js#L4) perhaps)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,5 +58,8 @@
       "eslint --fix",
       "prettier --write"
     ]
+  },
+  "dependencies": {
+    "glob": "^10.2.3"
   }
 }
--- a/scripts/lint-multi-repo.js
+++ b/scripts/lint-multi-repo.js
@@ -1,0 +1,58 @@
+#!/usr/bin/env node
+/* eslint-disable no-console */
+
+//internal
+const Logger = require('../packages/logger/lib/index');
+const eslintConfig = require('../packages/eslint-config/lib/index');
+
+//external
+const { glob } = require('glob');
+const { ESLint } = require('eslint');
+const fs = require('fs');
+const path = require('node:path');
+
+(async () => {
+	try {
+		// create an instance that extends the dotcom-reliability-kit/eslint-config via the imported module
+		const eslint = new ESLint({
+			useEslintrc: false,
+			// @ts-ignore >> not able to fix this and doesn't stop the script from running fine
+			overrideConfig: eslintConfig
+		});
+
+		// specify folderpath to lint
+		const folderPath = path.join(
+			__dirname,
+			'..',
+			'..',
+			'clone-all-ft-cp-repos',
+			'repos'
+		);
+
+		// fs.readdirSync returns an array of folderPaths, so filtering out unwanted folderpaths or it'll throw an error
+		const ignoreFolderList = ['.DS_Store'];
+
+		const arrayOfFolderPaths = fs
+			.readdirSync(folderPath)
+			.filter((folder) => !ignoreFolderList.includes(folder));
+
+		// now we have an array of folders to loop through...
+		for (const folder of arrayOfFolderPaths) {
+			// ... and only store the files that have a glob path with files ending in *.js...
+			const jsfiles = await glob(`${folderPath}/${folder}/**/*.js`);
+
+			// ... and then lint these *.js files...
+			const results = await eslint.lintFiles(jsfiles);
+
+			// ... and then format the results...
+			const formatter = await eslint.loadFormatter('stylish');
+			const resultText = formatter.format(results);
+
+			// ... and finally, output the results via the console
+			console.log(resultText);
+		}
+	} catch (error) {
+		process.exitCode = 1;
+		Logger.error(error);
+	}
+})();
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,9 @@
         "resources/*",
         "packages/*"
       ],
+      "dependencies": {
+        "glob": "^10.2.3"
+      },
       "devDependencies": {
         "@babel/core": "^7.21.8",
         "@babel/eslint-parser": "^7.21.8",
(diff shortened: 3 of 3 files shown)


alexmuller (commented):
This is really cool. One thing that I think we could improve is making the file path a bit more resilient.

Right now if you run this script from the `dotcom-reliability-kit/` directory it works great, but if you `cd scripts/` and run `node lint-multi-repo.js` it won't - because the `..` is relative to where you're running it from.

Changing the call to this:

```
eslint.lintFiles([`${__dirname}/../../**/*.js`])
```

makes it not dependent against where you're running it from, because `__dirname` is always the same.

And going even further using `path` will make the code more reusable:

```js
const path = require('node:path')
// etc
eslint.lintFiles([path.join(__dirname, '..', '..', '**', '*.js')])
```

---

It's also worth bearing in mind that this will lint all files in the directories at the same level as `dotcom-reliability-kit/` which I think is a good start, but that does include files like client-side JS which might cause a lot of errors.

CyntiBinti:
> This is really cool. One thing that I think we could improve is making the file path a bit more resilient.
> 
> Right now if you run this script from the `dotcom-reliability-kit/` directory it works great, but if you `cd scripts/` and run `node lint-multi-repo.js` it won't - because the `..` is relative to where you're running it from.
> 
> Changing the call to this:
> 
> ```
> eslint.lintFiles([`${__dirname}/../../**/*.js`])
> ```
> 
> makes it not dependent against where you're running it from, because `__dirname` is always the same.
> 
> And going even further using `path` will make the code more reusable:
> 
> ```js
> const path = require('node:path')
> // etc
> eslint.lintFiles([path.join(__dirname, '..', '..', '**', '*.js')])
> ```
> 
> It's also worth bearing in mind that this will lint all files in the directories at the same level as `dotcom-reliability-kit/` which I think is a good start, but that does include files like client-side JS which might cause a lot of errors.

Ohh thanks for this @alexmuller this is a very good point as better to be dynamic than hard-coded. Will def factor this in once I've figured out how best to lint the folders iteratively 😸

CyntiBinti:
> Current problem I'm trying to solve: Memory heap issue that occurs when I run the script. When looping through each folder, node.js isn't releasing memory after linting each folder/outputting the results to the terminal, and therefore it still runs out of memory eventually 😭 It gets as far as folder 109 out of 403 folders in the directory `'../clone-all-ft-cp-repos/repos/Financial-Times/'`. Setting the `NODE_OPTIONS=--max-old-space-size=1024000` also has not worked.
> 
> image

This is fine now as input size reduced from 500+ repos (all CP repo's) to only a necessary 126 repos (service tier bronze - platinum)

rowanmanning:
It's not directly equivalent but trying to work out what the difference is between this and your script, because the following works for me when run across all 500 repos. I wonder if the `--no-eslint` + `--config` bit is important because it's telling ESLint to not try and load config files from each of the directories which probs frees up a lot of memory:

```
NODE_OPTIONS=--max_old_space_size=8000 eslint --no-eslintrc --config <CONFIG_FILE> ./<REPOS_FOLDER>
```

scripts/lint-multi-repo.js
```diff
@@ -0,0 +1,50 @@
...
+const { ESLint } = require('eslint');
+const fs = require('fs');
+
+(async () => {
+	try {
+		// create an instance that extends the dotcom-reliability-kit/eslint-config via the imported module
+		const eslint = new ESLint({
+			useEslintrc: false,
+			// @ts-ignore >> not able to fix this and doesn't stop the script from running fine
+			overrideConfig: eslintConfig
```
rowanmanning, line 19: OK maybe try this?

```suggestion
			overrideConfig: eslintConfig,
			useEslintrc: false,
```

CyntiBinti: But that's what I thought by assigning this `useEslintrc` property to `false` meant it was not using the native eslint files 🤔 . Seems you feel by changing the order this should work now?

rowanmanning: Ignore me entirely 😂 I somehow missed that you were already setting this property

CyntiBinti:
> 

So odd! When I run your command `NODE_OPTIONS=--max_old_space_size=8000 eslint --no-eslintrc --config /Users/cynthia.mbulu/Documents/code/dotcom-reliability-kit/packages/eslint-config/lib/index.js ./repos-old` it outputs the linted results for all 500+ repos in CP 😮 : **_✖ 56865 problems (56862 errors, 3 warnings)_**

Maybe you're right, and the ESLint class does some extra optimisation under the hood that slows it down in my script 🤷🏽‍♀️

CyntiBinti:
> 

@alexmuller would this mean that a user would need to clone dotcom-reliability-kit/ at the same level of the directory they'd want to lint? 
else it would just lint every folder at the same level of dotcom-reliability-kit/ (unwanted behaviour)

CyntiBinti:
> This is really cool. One thing that I think we could improve is making the file path a bit more resilient.
> 
> Right now if you run this script from the `dotcom-reliability-kit/` directory it works great, but if you `cd scripts/` and run `node lint-multi-repo.js` it won't - because the `..` is relative to where you're running it from.
> 
> Changing the call to this:
> 
> ```
> eslint.lintFiles([`${__dirname}/../../**/*.js`])
> ```
> 
> makes it not dependent against where you're running it from, because `__dirname` is always the same.
> 
> And going even further using `path` will make the code more reusable:
> 
> ```js
> const path = require('node:path')
> // etc
> eslint.lintFiles([path.join(__dirname, '..', '..', '**', '*.js')])
> ```
> 
> It's also worth bearing in mind that this will lint all files in the directories at the same level as `dotcom-reliability-kit/` which I think is a good start, but that does include files like client-side JS which might cause a lot of errors.

@alexmuller would this mean that a user would need to clone dotcom-reliability-kit/ at the same level of the directory they'd want to lint?
else it would just lint every folder at the same level of dotcom-reliability-kit/ (unwanted behaviour)

scripts/lint-multi-repo.js
```diff
@@ -0,0 +1,55 @@
...
+			'..',
+			'..',
+			'clone-all-ft-cp-repos',
+			'repos'
+		);
+		// fs.readdirSync returns an array of folderPaths, so filtering out unwanted folderpaths or it'll throw an error
+		const arrayOfFolderPaths = fs
+			.readdirSync(folderPath)
+			.filter((folder) => folder !== '.DS_Store');
+
```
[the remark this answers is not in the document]
CyntiBinti: Great idea! thank you for this will update my code 😁

package.json
```diff
@@ -58,5 +58,8 @@
       "eslint --fix",
       "prettier --write"
     ]
+  },
+  "dependencies": {
+    "glob": "^10.2.3"
```
rowanmanning, line 63: **question:** is it possible to get by without this dependency in the root of the project? It's not a super big one but if it's at the root then it'll be installed for every build of Reliability Kit regardless of whether we're actually running the script or not, which is a little bit of an overhead.

rowanmanning (commented):
One small comment but may be irrelevant - is this ready for review as in it'll live within Reliability Kit? Or is the intention for it to move to the new `cp-reliability-scripts` repo?

PR #576, opened 2023-05-15
chore: use the new ESLint config

rowanmanning:
This uncovered one issue in some of our tests where we were setting a duplicate key.


CyntiBinti (approved):
🙌🏽

PR #582, opened 2023-05-23
feat!: add new rules from eslint-config-next

rowanmanning:
This adds new ESLint rules which are defined in eslint-config-next. It's fine to duplicate these rules across configs and these can result in bugs. Unfortunately this means releasing a v2.0.0 pretty quickly but I think that's fine - let's not shy away from adding new rules!


CyntiBinti (commented):
Suggestion: Oh wait - what about?:
- [no-irregular-whites](https://eslint.org/docs/latest/rules/no-irregular-whitespace)
- no-console

rowanmanning:
Ah yeah I can get behind `no-irregular-whitespace` but not sure whether `no-console` is the remit of Reliability Kit – in a script or something it's probably valid to use `console.log`?

alexmuller:
Woop!

PR #584, opened 2023-05-23
chore: add type checking to the ESLint config

rowanmanning:
Because the ESLint config isn't in a file named `.eslintrc` or similar, it doesn't get type-hinting in VSCode and we're also not certain that the configuration is valid.

We already run type checking against the code with TypeScript, so we can validate our ESLint config the same way.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2274,6 +2274,22 @@
         "@types/node": "*"
       }
     },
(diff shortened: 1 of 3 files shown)


CyntiBinti (approved):
Good shout - thank you!

PR #603, opened 2023-06-12
add request headers seriealization

### LINKED ISSUE #277: Support serialising iterable request headers
The `serializeRequest` method currently expects `request.headers` to always be an object. There are several contexts where the headers can be an iterable which it could be good to support. I chatted about this with @apaleslimghost.

In [Apollo Server's `requestDidStart`](https://www.apollographql.com/docs/apollo-server/v3/integrations/plugins-event-reference/#requestdidstart) and also the [`Headers` object](https://developer.mozilla.org/en-US/docs/Web/API/Headers) used by the `fetch` API, the HTTP headers are not an object, they're an iterable. Currently this means you need to do the following if you want to serialise headers for these types of request:

```js
// Assuming `request` is a `Request` instance as provided by the fetch API
serializeRequest({
    ...request,
    headers: Object.fromEntries(Array.from(request.headers))
});
```

## Ideal solution

It might be good to support iterables in the serialiser itself. We'd have to update the types and then change the code to check whether the value we're given is iterable. We should be able to test for this with [`Symbol.iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator).

## Alternatives

We don't do this, and keep iterable header serialisation in user code at the moment. We currently only have one use-case for this in the Customer Products API rationalisation team so maybe we revisit if we find that more people need it (or adoption of Apollo increases).

Nazehs:
In certain contexts like Apollo Server's requestDidStart and the Headers object used by the fetch API, HTTP headers are represented as iterables rather than objects. As a result, when serializing headers for these types of requests, extra code is required to convert the iterable headers into an object.

This PR update the serializeRequest method to support iterable headers directly. The necessary changes include updating the types and modifying the code to check if the provided headers are iterable using Symbol.iterator. If the headers are iterable, they will be processed accordingly. This improvement will simplify the serialization process and eliminate the need for users to handle iterable headers in their own code.

Resolves #277

### DIFF
--- a/packages/serialize-request/lib/index.js
+++ b/packages/serialize-request/lib/index.js
@@ -8,7 +8,7 @@
 
 /**
  * @typedef {object} BasicRequest
- * @property {Object<string, string>} [headers]
+ * @property {Headers | Object<string, string> | Iterable<[string, string]>} [headers]
  *     The request HTTP headers.
  * @property {string} [method]
  *     The request HTTP method.
@@ -127,11 +127,13 @@ function serializeRequest(request, options = {}) {
 	}
 
 	// Serialize the headers
-	if (
-		request.headers &&
+	const headersIsObject =
 		typeof request.headers === 'object' &&
 		!Array.isArray(request.headers) &&
-		request.headers !== null
+		request.headers !== null;
+	if (
+		request.headers &&
+		(headersIsObject || isIterableHeadersObject(request.headers))
 	) {
 		requestProperties.headers = serializeHeaders(
 			request.headers,
@@ -154,7 +156,7 @@ function serializeRequest(request, options = {}) {
  * Serialize request headers.
  *
  * @private
- * @param {Record<string, any>} headers
+ * @param {Headers | Record<string, any> | Iterable<[string, string]>} headers
  *     The headers object to serialize.
  * @param {Array<string>} includeHeaders
  *     An array of request headers to include.
@@ -162,8 +164,20 @@ function serializeRequest(request, options = {}) {
  *     Returns the serialized request headers.
  */
 function serializeHeaders(headers, includeHeaders) {
+	const headersObject = {};
+	let iterableHeaders =
+		Array.isArray(headers) || isIterableHeadersObject(headers)
+			? headers
+			: Object.entries(headers);
+
+	for (const [header, value] of iterableHeaders) {
+		if (typeof header === 'string' && typeof value === 'string') {
+			headersObject[header.toLowerCase()] = value;
+		}
+	}
+
 	return Object.fromEntries(
-		Object.entries(headers).filter(([header]) =>
+		Object.entries(headersObject).filter(([header]) =>
 			includeHeaders.includes(header)
 		)
 	);
@@ -189,6 +203,16 @@ function createSerializedRequest(properties) {
 		},
 		properties
 	);
+}
+
+/**
+ * @param {any} value
+ *     The value to test.
+ * @returns {value is Iterable<[string, string]>}
+ *     Returns whether a value is iterable.
+ */
+function isIterableHeadersObject(value) {
+	return value && typeof value?.[Symbol.iterator] === 'function';
 }
 
 module.exports = serializeRequest;
--- a/packages/serialize-request/test/unit/lib/index.spec.js
+++ b/packages/serialize-request/test/unit/lib/index.spec.js
@@ -220,6 +220,33 @@ describe('@dotcom-reliability-kit/serialize-request', () => {
 				});
 			});
 		});
+		describe('when `request.headers` is iterable', () => {
+			beforeEach(() => {
+				request.headers = new Map([
+					['Content-Type', 'application/json'],
+					['Authorization', 'Bearer token']
+				]);
+			});
+			it('should handle headers as an iterable', () => {
+				expect(serializeRequest(request)).toMatchObject({
+					headers: { 'content-type': 'application/json' }
+				});
+			});
+		});
+
+		describe('when `request.headers` is iterable but has non-string values', () => {
+			beforeEach(() => {
+				request.headers = new Map([
+					['Content-Type', 123],
+					['accept', {}]
+				]);
+			});
+			it('should ignore the headers with non-string values', () => {
+				expect(serializeRequest(request)).toMatchObject({
+					headers: {}
+				});
+			});
+		});
 
 		describe('when `request.route.path` is not a string', () => {
 			beforeEach(() => {


packages/serialize-request/lib/index.js
```diff
@@ -154,16 +154,33 @@ function serializeRequest(request, options = {}) {
...
+
+	if (Array.isArray(headers)) {
+		for (const [header, value] of headers) {
+			headersObject[header.toLowerCase()] = value;
+		}
+	} else if (typeof headers[Symbol.iterator] === 'function') {
+		// @ts-ignore
+		for (const [header, value] of headers) {
+			headersObject[header.toLowerCase()] = value;
+		}
```
rowanmanning, line 175: **suggestion:** it's a little more code but we can remove this `@ts-ignore` by defining a type checking function and using that in this condition. This will make sure we know we're being type-safe. If you change this line to something like:

```suggestion
	} else if (isIterableHeadersObject(headers)) {
		// @ts-ignore
		for (const [header, value] of headers) {
			headersObject[header.toLowerCase()] = value;
		}
```

Then we can define our type checking function like this and TypeScript will understand that, within that condition, the value has an iterator:

```js
/**
 * @param {any} value
 *     The value to test.
 * @returns {value is Iterable<string, string>}
 *     Returns whether a value is iterable.
 */
function isIterableHeadersObject(value) {
	return typeof value?.[Symbol.iterator] === 'function';
}
```

packages/serialize-request/lib/index.js
```diff
@@ -154,16 +154,33 @@ function serializeRequest(request, options = {}) {
...
  *     The headers object to serialize.
  * @param {Array<string>} includeHeaders
  *     An array of request headers to include.
  * @returns {SerializedRequestHeaders}
  *     Returns the serialized request headers.
  */
 function serializeHeaders(headers, includeHeaders) {
+	const headersObject = {};
+
+	if (Array.isArray(headers)) {
```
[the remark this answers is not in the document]
rowanmanning: Some clarification on this one, by type guarding I mean making sure that the headers we get are _actually_ strings. TypeScript will check this as a kind of linting step but there's nothing stopping someone passing in malformed data. It's safer if we do something like the following (feel free to reduce repetition / tweak obvs):

```js
if (Array.isArray(headers)) {
	for (const [header, value] of headers) {
		if (typeof header === 'string' && typeof value === 'string') {
			headersObject[header.toLowerCase()] = value;
		}
	}
} else if (isIterableHeadersObject(headers)) {
	for (const [header, value] of headers) {
		if (typeof header === 'string' && typeof value === 'string') {
			headersObject[header.toLowerCase()] = value;
		}
	}
} else {
	for (const header of Object.keys(headers)) {
		if (typeof header === 'string' && typeof value === 'string') {
			headersObject[header.toLowerCase()] = value;
		}
	}
}
```

Perhaps a less repetitive way of writing this would be something like:

```js
let iterableHeaders = [];

if (Array.isArray(headers) || isIterableHeadersObject(headers)) {
	iterableHeaders = headers;
} else {
	iterableHeaders = Object.keys(headers);
}

for (const [header, value] of iterableHeaders) {
	if (typeof header === 'string' && typeof value === 'string') {
		headersObject[header.toLowerCase()] = value;
	}
}
```

packages/serialize-request/lib/index.js
```diff
@@ -191,6 +204,16 @@ function createSerializedRequest(properties) {
 	);
 }
 
+/**
+ * @param {any} value
+ *     The value to test.
+ * @returns {value is Iterable<string, string>}
```
rowanmanning, line 210: **suggestion:** I think this needs to change to match the value in the function above to work properly, I got some TS issues when testing and this fixed it:

```suggestion
 * @returns {value is Iterable<[string, string]>}
```

packages/serialize-request/lib/index.js
```diff
@@ -128,9 +128,10 @@ function serializeRequest(request, options = {}) {
...
 	// Serialize the headers
 	if (
-		request.headers &&
-		typeof request.headers === 'object' &&
-		!Array.isArray(request.headers) &&
+		((request.headers &&
+			typeof request.headers === 'object' &&
+			!Array.isArray(request.headers)) ||
+			isIterableHeadersObject(request.headers)) &&
 		request.headers !== null
```
rowanmanning, line 135: **suggestion:** this works as expected but it took me a few tries to read it, could we maybe make it a little more readable by using a variable? Something like:

```js
// Serialize the headers
const headersIsObject =
	typeof request.headers === 'object' &&
	!Array.isArray(request.headers) &&
	request.headers !== null;
if (
	request.headers &&
	(headersIsObject || isIterableHeadersObject(request.headers))
) {
	requestProperties.headers = serializeHeaders(
		request.headers,
		includeHeaders
	);
}
```

packages/serialize-request/lib/index.js
```diff
@@ -154,16 +155,28 @@ function serializeRequest(request, options = {}) {
  * Serialize request headers.
  *
  * @private
- * @param {Record<string, any>} headers
+ * @param {Headers | Record<string, any> | Iterable<[string, string]>} headers
```
rowanmanning, line 158: **issue:** line 11 of this file will also need to change to include the new header types, otherwise I get a Type error when I try to pass an iterable in. Updating the `BasicRequest` type to this works for me:

```js
/**
 * @typedef {object} BasicRequest
 * @property {Headers | Object<string, string> | Iterable<[string, string]>} [headers]
 *     The request HTTP headers.
 * @property {string} [method]
 *     The request HTTP method.
 * @property {string} [url]
 *     The request URL.
 */
```

packages/serialize-request/test/unit/lib/index.spec.js
```diff
@@ -220,6 +220,19 @@ describe('@dotcom-reliability-kit/serialize-request', () => {
...
+					['Content-Type', 'application/json'],
+					['Authorization', 'Bearer token']
+				]);
+			});
+			it('should handle headers as an iterable', () => {
+				expect(serializeRequest(request)).toMatchObject({
+					headers: { 'content-type': 'application/json' }
+				});
+			});
+		});
```
rowanmanning, line 235: **suggestion:** this test is great 🙂 I wonder whether we should also be certain about the string checking to make sure this can't regress? Maybe another test like:

```suggestion
		});

		describe('when `request.headers` is iterable but has non-string values', () => {
			beforeEach(() => {
				request.headers = new Map([
					['Content-Type', 123],
					['accept', {}]
				]);
			});
			it('should ignore the headers with non-string values', () => {
				expect(serializeRequest(request)).toMatchObject({
					headers: {}
				});
			});
		});
```

rowanmanning (changes requested):
This is looking really great, thanks @Nazehs! It's super close, really happy with most of the code but I'm having a few type issues when trying to use it. I've made comments explaining the changes that fix it as well as a suggested test.

packages/serialize-request/lib/index.js
rowanmanning: Just this one to go @Nazehs, everything else looking perfect. Sorry the line number is wrong – I can't add a suggestion for a line that wasn't edited in the PR but this needs to go near line 11

rowanmanning (approved):
Amazing! Thanks so much for your work on this, let's get it merged

PR #612, opened 2023-06-23
set require-atomic-updates to warn level

CyntiBinti:
This sets require-atomic-updates to a warn level as we were seeing it error as a false positive. 
Linked to [JIRA-629](https://financialtimes.atlassian.net/browse/CPREL-629)

### DIFF
--- a/packages/eslint-config/README.md
+++ b/packages/eslint-config/README.md
@@ -30,10 +30,10 @@ Avoid using .eslintrc file format (without a file extension), as [this has been 
 Add the following command to the `scripts` object in your repository's package.json:
 
 ```
-"lint:js": "eslint '**/*.{js,jsx,json,yml,yaml}'"
+"lint:js": "eslint ."
 ```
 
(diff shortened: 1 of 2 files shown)


rowanmanning (approved):
Nice 👍

rowanmanning:
Sorry I reviewed without looking at the commit messages :) the second one should be either a `fix` or `feat` as `chore` doesn't trigger a release. The one we choose depends on whether we want a [SemVer](https://semver.org/) patch or minor release, I don't think I have strong opinions on which

CyntiBinti:
> Sorry I reviewed without looking at the commit messages :) the second one should be either a `fix` or `feat` as `chore` doesn't trigger a release. The one we choose depends on whether we want a [SemVer](https://semver.org/) patch or minor release, I don't think I have strong opinions on which

Ahh cool - will bear in mind for next time and change this one to fix. Thanks! 😺

PR #621, opened 2023-07-06
fix: address logging infinite recursion issue

rowanmanning:
This cropped up as a bug recently where Content Discovery were trying to log an entire Request object, which references itself. This resulted in Node.js running out of memory because we were infinitely recursing over the properties.

This is a bug that's always been present in n-mask-logger.

I've addressed the issue with my first legit use of `WeakSet` and I'm quite pleased with myself about it 😁


CyntiBinti (approved):
Yay love a good hash set 🙌🏽

PR #623, opened 2023-07-07
fix: properly clone error instances

### LINKED ISSUE #611: The logger strips important error information when errors are passed as properties
When we clone the log data (in order to stop original objects being modified) the library we're using strips important error information. This means that our recent feature (allowing error objects to be passed in as properties) is broken.

Thanks @metart43 for spotting this!

**Node.js version:** All
**Impacted package(s):** Logger

## Steps to reproduce

I wrote a test case (see PR #610) to illustrate the problem. If you want to reproduce the issue:

```js
const { OperationalError } = require('@dotcom-reliability-kit/errors');

// Works as expected, logs {error: { name: 'OperationalError', code: 'code1', message: 'msg1', ... }}
logger.info(new OperationalError({
    code: 'code1',
    message: 'msg1'
}));

// Strips information, logs {error: { name: 'Error', code: 'UNKNOWN', message: 'msg2', ... }}
logger.info({
    error: new OperationalError({
        code: 'code2',
        message: 'msg2'
    })
})
```

## Expected behaviour

The error should be logged with all details (`name`, `code`, etc). I have a pretty strong suspicion that this just won't work with the current cloning library, perhaps it won't work with any cloning library. We maybe need to assess whether it's that important to not modify the original log object.

rowanmanning:
This switches out the library we use for cloning objects because the original one did not properly handle error objects, which resulted in a loss of log data. This works with Lodash as indicated by the now-passing test - we should now not lose error info.

This fixes an issue Artem was having when trying to use the `error` property of log data:

```js
logger.error({
    error: new Error('...')
})
```

Resolves #611. Closes #610.

### DIFF
--- a/packages/logger/lib/logger.js
+++ b/packages/logger/lib/logger.js
@@ -1,6 +1,6 @@
 const pino = require('pino').default;
 const serializeError = require('@dotcom-reliability-kit/serialize-error');
-const { default: clone } = require('@ungap/structured-clone');
+const clone = require('lodash.clonedeep');
 const appInfo = require('@dotcom-reliability-kit/app-info');
 
 /**
@@ -146,7 +146,7 @@ class Logger {
 	/**
 	 * Create a logger.
 	 *
-	 * @param {LoggerOptions & PrivateLoggerOptions} [options = {}]
+	 * @param {LoggerOptions & PrivateLoggerOptions} [options]
 	 *     Options to configure the logger.
 	 */
 	constructor(options = {}) {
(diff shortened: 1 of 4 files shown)


packages/logger/lib/logger.js
```diff
@@ -1,6 +1,6 @@
 const pino = require('pino').default;
 const serializeError = require('@dotcom-reliability-kit/serialize-error');
-const { default: clone } = require('@ungap/structured-clone');
+const clone = require('lodash.clonedeep');
```
metart43, line 3: Sorry for being annoying, but I'd require `cloneDeep`. It seems like lodash has a separate method called `clone`

```suggestion
const cloneDeep = require('lodash.clonedeep');
```

metart43:
Thanks for resolving this @rowanmanning !

packages/logger/lib/logger.js
rowanmanning: I think I'm OK with it staying as-is, as it results in an easier-to-read diff and the library we use to clone is kind of an implementation detail. We don't install lodash.clone so don't think that'd cause confusion, and we also didn't use the name `structuredClone` when we were using that library

metart43 (approved):

CyntiBinti (approved):
🎉

Issue #509, opened 2023-04-12
We now see ESLint warnings for some JSDoc type annotations


rowanmanning:
Since #507 we see a lot of ESLint warnings for our object types. E.g.

```
warning  Use object shorthand or index signatures instead of `Object`, e.g., `{[key: string]: string}`
```

This is because of a breaking change in [v41 of the plugin](https://github.com/gajus/eslint-plugin-jsdoc/releases/tag/v41.0.0) we use combined with the fact we enable the plugin's TypeScript mode.

**Node.js version:** all
**Impacted package(s):** all

## Steps to reproduce

1. Run `npm run lint`
2. See warnings

## Expected behaviour

I'd expect us not to have a bunch of fixable warnings. We should decide whether we want to disable TypeScript mode _or_ switch all our types to be more like TypeScript. If we decide the latter then we'd need to change things in the JSDoc like:

```
Object<string, string>
```

to:

```
{[key: string]: string}
```

If we were doing this then I'd probably suggest also changing our array notation from this:

```
Array<string>
```

to this:

```
string[]
```

rowanmanning:
I'd vote for option 2: making our existing Types more TypeScript-looking as I think the syntax is better-understood than JSDoc

CyntiBinti:
This has now been fixed :)

PR #647, opened 2023-07-18
chore: eslint warnings for object type defs

CyntiBinti:
This PR fixes the ESlint warnings for JSDoc type checking (issue #509). All the 25 warnings have been fixed 🎉 

For the `Iterable` type warning problem, this was fixed based on [github issue help](https://github.com/gajus/eslint-plugin-jsdoc/issues/280)

[JIRA](https://financialtimes.atlassian.net/browse/CPREL-650)

image

### DIFF
--- a/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
+++ b/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
@@ -3,7 +3,7 @@
  *
  * @param {string} logString
  *     The log string to parse.
- * @returns {Array<{[key: string]: any}>}
+ * @returns {{[key: string]: any}[]}
  *     Returns the parsed JSON logs as an array of log objects.
  */
 function splitAndParseJsonLogs(logString) {
--- a/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
+++ b/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
@@ -3,7 +3,7 @@
  *
  * @param {string} logString
  *     The log string to parse.
- * @returns {Array<Object<string, any>>}
+ * @returns {Array<{[key: string]: any}>}
  *     Returns the parsed JSON logs as an array of log objects.
  */
 function splitAndParseJsonLogs(logString) {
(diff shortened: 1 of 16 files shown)


packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
```diff
@@ -3,7 +3,7 @@
  *
  * @param {string} logString
  *     The log string to parse.
- * @returns {Array<Object<string, any>>}
+ * @returns {Array<{[key: string]: any}>}
```
rowanmanning, line 6: **suggestion (non-blocking):** we could also look at adjusting array syntax as well. It's not causing warnings but it makes sure we're always using TypeScript-notation rather than JSDoc. This is a more complex example but think this works:

```suggestion
 * @returns {{[key: string]: any}[]}
```

A simpler example would be:

```diff
- @returns {Array<string>}
+ @returns {string[]}
```

I think this should be findable by searching for all instances of `Array<`.

rowanmanning (commented):
One small suggestion, but if you'd rather open as a separate issue then that's fine

packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
CyntiBinti: Thanks for this - I've also changed the Array typedefs to be more like TypeScript :)

rowanmanning (commented):
Looking great, thanks for doing this! I just have one change I didn't spot the first time around. Because this is a `fix:` level commit this will cause new releases to be tagged up for all packages, however there's no actual benefit to end-users of Reliability Kit, nothing changes about the way they use the packages.

For this reason I think all the commits should be `chore:`-level, which is more appropriate if we're doing internal tidying with no actual impact. Does that make sense?

CyntiBinti:
> Looking great, thanks for doing this! I just have one change I didn't spot the first time around. Because this is a `fix:` level commit this will cause new releases to be tagged up for all packages, however there's no actual benefit to end-users of Reliability Kit, nothing changes about the way they use the packages.
> 
> For this reason I think all the commits should be `chore:`-level, which is more appropriate if we're doing internal tidying with no actual impact. Does that make sense?

Yeah this makes total sense, no point to cause unnecessary bumps. Let me rebase and change the commit message of the commits 😌

rowanmanning (approved):
🚀🚀🚀

PR #648, opened 2023-07-19
feat: add a BaseError class

### LINKED ISSUE #238: Provide non-operational error classes
Make it possible to easily throw non-operational errors with additional information, e.g. error `code`.

In our documentation, [we advise against using `OperationalError` in the context of a library](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/getting-started/throwing-errors.md#operational-errors-in-library-code). This is because a library cannot know whether the application it's running in actually understands how to handle that error – it's not possible to confirm that the error is operational. However we _do_ want errors from our libraries to still include useful information, such as `code` or `relatesToSystems`.

## Ideal solution

I think the ideal solution would be to provide a `BaseError` class in @dotcom-reliability-kit/errors which provides all the same functionality of `OperationalError` except that it has the `isOperational` property set to `false`. Most of the functionality from our `OperationalError` class could then move to this and `OperationalError` would just default `isOperational` to `true`.

We might also want to change the way our classes inherit from each other because of this. E.g. do we want to support a non-operational `UpstreamServiceError` for use in libraries?

## Alternatives

* We could do nothing, but I suspect this will mean that library code _will_ start to introduce operational errors. Do we care that much if we're getting more error information in Splunk? Maybe not.

rowanmanning:
Reliability Kit provides a neat way to set additional properties on error objects during construction, but all the errors we provide are marked as operational. This means that Reliability Kit errors ideally shouldn't be used in library code and we lose the convenience of our extended constructors.

This PR adds in a `BaseError` class which is marked as _not_ operational. The `OperationalError` class now extends this so we're not duplicating logic.

This resolves #238.

See https://github.com/Financial-Times/cp-content-pipeline/pull/423 for a real world use-case.


packages/errors/README.md
```diff
@@ -200,6 +201,25 @@ throw new UserInputError({
 });
 ```
 
+### `BaseError`
+
+Our getting started guide talks about [Operational vs Programmer errors](../../docs/getting-started/throwing-errors.md#types-of-error). The rest of the errors in this library provide operational errors, but they all extend a base class named `BaseError`. This class offers up some of the conveniences of `OperationalError` but it's marked as non-operational:
```
apaleslimghost, line 206: **suggestion**: in the world where BaseError existed from the start, it wouldn't make sense to talk about it "offering up some of the convenience of `OperationalError`". i think it would be clearer to new users of the library to talk about what it does in concrete terms, then mention that the other error classes inherit that.

packages/errors/lib/base-error.js
```diff
@@ -0,0 +1,170 @@
...
+			this.code = BaseError.normalizeErrorCode(data.code);
+		}
+
+		if (data.cause instanceof Error) {
+			this.cause = data.cause;
+		}
+
+		for (const [key, value] of Object.entries(data)) {
+			// @ts-ignore TypeScript does not properly infer the constructor
+			if (!this.constructor.reservedKeys.includes(key)) {
```
apaleslimghost, line 108: **thought**: is there a way to do typecasting with JSDoc? in Typescript i'd do `(this.constructor as typeof BaseError)`

apaleslimghost (approved):
ily

packages/errors/README.md
rowanmanning: I did start writing it like that but I think really we don't want _loads_ of use of this? I worry slightly if all the docs for the properties are on `BaseError` then people will use that in places where `OperationalError` is more correct. It's mostly gonna be library authors who need this so I kinda wanted it to fly under the radar a little for now, alright if I make another ticket to review the docs in future and leave this as-is?

packages/errors/lib/base-error.js
rowanmanning: OMG it is! I'm gonna go and update all of the other ts-ignores for this issue [0afacc8](https://github.com/Financial-Times/dotcom-reliability-kit/pull/648/commits/0afacc8f631c0b58cc4cd1eb4b9995d65c65cb81)

packages/errors/README.md
apaleslimghost: that's fair

packages/errors/lib/base-error.js
```diff
@@ -0,0 +1,170 @@
...
+		return code
+			.trim()
+			.toUpperCase()
+			.replace(/[^a-z0-9_]+/gi, '_');
+	}
+}
+
+module.exports = BaseError;
+
+module.exports.default = module.exports;
```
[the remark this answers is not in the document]
rowanmanning: We're doing this in all the Reliability Kit modules for compatibility with the different module systems we use. In certain cases a TypeScript project will compile this:

```js
import BaseError from '@dotcom-reliability-kit/errors/lib/base-error';
```

to the following CommonJS:

```js
const BaseError = require('@dotcom-reliability-kit/errors/lib/base-error').default;
```

We don't really have any control over this, and aliasing the `default` export like this helps us avoid having to compile the code as part of a build step.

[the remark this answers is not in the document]
rowanmanning: Oh and in regular CommonJS land, we don't want people to have to specify `default`, so we change `module.exports` to allow users to do this:

```js
const BaseError = require('@dotcom-reliability-kit/errors/lib/base-error');
```

rather than having to remember to add `.default` to the end.

CyntiBinti (approved):
🙌🏽

packages/errors/lib/base-error.js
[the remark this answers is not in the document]
CyntiBinti: Ah I seeee, ok. That helps to understand. Thank you 😸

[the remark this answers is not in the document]
CyntiBinti: So for me to bear in mind, if we're building something (e.g. a new package) that may be used by different module systems, it's preferred to double export like this to ensure it's compatible once installed?

PR #651, opened 2023-07-19
chore: attempt to replicate Jest logger exit issue

rowanmanning:
I'm trying to replicate the issue that Artem reported on Slack about the logger not allowing Jest to exit the process. So far I haven't had any luck but maybe I've reduced the test case down too much.

To try this locally checkout this branch and run:

```
jest packages/logger/test/end-to-end/express.spec.js
```

---

Note: the Node.js 16 test failing is unrelated – I'm using native fetch which is only available in Node.js 18+

### DIFF
--- a/packages/logger/test/end-to-end/express.spec.js
+++ b/packages/logger/test/end-to-end/express.spec.js
@@ -1,0 +1,34 @@
+const express = require('express');
+const logger = require('../..');
+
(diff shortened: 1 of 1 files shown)


rowanmanning:
We worked it out, and we'll be solving with documentation on Jest mocking in #693.

PR #653, opened 2023-07-20
fix: avoid double-registering Crash Handler

rowanmanning:
If Crash Handler is registered twice _or_ it's registered alongside another uncaughtException handler then it'll behave in an unexpected way. It's best if we just avoid registering in these scenarios.

Because we want to know if Crash Handler isn't being registered for an app, we're logging a warning with an error code. I'll set up a Splunk alert so that we can see immediately if an app breaks their crash handling.


CyntiBinti (approved):
Looking gooood 🙌🏽

PR #660, opened 2023-08-04
feat!: switch to Reliability Kit logger

rowanmanning:
This removes n-logger from all packages, switching it to Reliability Kit logger. To make sure that teams can undo this temporarily, I've added the ability to configure the logger to the only module where it wasn't possible - middleware-render-error-info.

This means that these modules will no longer log to Splunk unless the app they're used in is configured to forward `stdout` logs there.

n-logger is still installed as a development dependency in a few places, mostly for testing compatibility with Reliability Kit logger and also in places where we're importing n-express for end-to-end middleware tests. You can verify that n-logger is not installed as a production dependency with this:

```sh
npm ls --depth 100 --omit dev @financial-times/n-logger
```

Part of [CPREL-668](https://financialtimes.atlassian.net/browse/CPREL-668).

[CPREL-668]: https://financialtimes.atlassian.net/browse/CPREL-668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CyntiBinti (approved):
🎉

PR #681, opened 2023-08-16
feat: add a fetch-error-handler package

rowanmanning:
This introduces a new package named @dotcom-reliability-kit/fetch-error-handler. It's designed to reduce boilerplate when handling `fetch` errors and provide better debugging information when fetches fail.

There was discussion about this feature in a [discussion on GitHub here](https://github.com/Financial-Times/dotcom-reliability-kit/discussions/643). Big thanks to @AniaMakes, @phelipemaia, @vbachev, and @wheresrhys for their contributions.

There are a few key differences in the implementation:

  1. We decided not to catch network and timeout errors as a first pass. `fetch` already throws errors for these and they could be added in a future version without a breaking change.

  2. We decided not to try and handle JSON parsing of the body. Again we wanted to get an MVP out and then see how people use it before tackling invalid response bodies.

A copy of the usage guide is outlined below, this appears in the `README`. This will be released as `v0.1.0` to indicate that it's an experimental package. We'll test in a few Customer Products systems and gather feedback ahead of a `v1.0.0` release at some point in the near future.

---

## Usage

Install `@dotcom-reliability-kit/fetch-error-handler` as a dependency:

```bash
npm install --save @dotcom-reliability-kit/fetch-error-handler
```

Include in your code:

```js
import { handleFetchErrors } from '@dotcom-reliability-kit/fetch-error-handler';
// or
const { handleFetchErrors } = require('@dotcom-reliability-kit/fetch-error-handler');
```

You can use this function with any `fetch` call to throw appropriate errors based on the [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) that you get back.

There are many ways to use it, as long as it is `await`ed and is called with either a `Response` object or a `Promise` that resolves to a `Response`. The function is asynchronous and resolves with the `Response` that it was called with:

```js
// Pass the function into the `then` function of a `fetch`.
// Note: this must be `then` and not `catch`
const response = await fetch('https://httpbin.org/status/500').then(handleFetchErrors);

// Wrap the fetch function. You can do this safely without
// awaiting the fetch itself
const response = await handleFetchErrors(
    fetch('https://httpbin.org/status/500')
);

// Pass in a response manually:
const response = await fetch('https://httpbin.org/status/500');
await handleFetchErrors(response);
```

If the reponse `ok` property is `false`, i.e. when the status code is `400` or greater, then errors will be thrown.

### Errors thrown

We throw different errors depending on the status code we get back from the `fetch` call.

#### Client errors

If the URL you fetched responds with a status code in the range of `400–499` then this normally indicates that  something is wrong with the _current_ system. Maybe we're sending data in an invalid format or our API key is invalid. For this we throw a generic `500` error to indicate an issue with _our_ system. It'll be an [`HTTPError`](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#httperror). This error will have the following properties to help you debug:

```js
error.statusCode // 500
error.code // FETCH_CLIENT_ERROR
error.data.upstreamUrl // The URL that was fetched
error.data.upstreamStatusCode // The status code that the URL responded with
```

#### Server errors

If the URL you fetched responds with a status code in the range of `500–599` then this indicates something is wrong with the _upstream_ system. For this we can output an [`UpstreamServiceError`](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#httperror) and attribute it to the system we're fetching from. This error will have the following properties to help you debug:

```js
error.statusCode // 502
error.code // FETCH_SERVER_ERROR
error.data.upstreamUrl // The URL that was fetched
error.data.upstreamStatusCode // The status code that the URL responded with
```

#### Unknown errors

If the URL you fetched responds with an `ok` property of `false` and a status code outside of the `400–599` range, then it's unclear what's happened but we reject with an error anyway to make sure we're able to debug. We output an [`HTTPError`](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#httperror):

```js
error.statusCode // 500
error.code // FETCH_UNKNOWN_ERROR
error.data.upstreamUrl // The URL that was fetched
error.data.upstreamStatusCode // The status code that the URL responded with
```

### Creating your own handler

You can customise the thrown errors by creating your own fetch handler and passing in [options](#createfetcherrorhandler-configuration-options).

Include in your code:

```js
import { createFetchErrorHandler } from '@dotcom-reliability-kit/fetch-error-handler';
// or
const { createFetchErrorHandler } = require('@dotcom-reliability-kit/fetch-error-handler');
```

Create and use your own handler (the handler supports all the same usage methods as [outlined here](#usage)):

```js
const handleFetchErrors = createFetchErrorHandler({
    upstreamSystemCode: 'httpbin'
});
const response = await fetch('https://httpbin.org/status/500').then(handleFetchErrors);
```

If you want a custom handler just for one `fetch` call, then you can shorten the above example to:

```js
const response = await fetch('https://httpbin.org/status/500').then(createFetchErrorHandler({
    upstreamSystemCode: 'httpbin'
}))
```

### `createFetchErrorHandler` configuration options

Config options can be passed into the `createFetchErrorHandler` function to change the behaviour of the handler.

#### `options.upstreamSystemCode`

Attribute any fetch errors to a given [Biz Ops system](https://biz-ops.in.ft.com/list/Systems). This allows you to easily spot in your logs when an upstream system is the cause of an error. This must be an `String` and a valid [system code](https://tech.in.ft.com/tech-topics/operability/biz-ops/unique-identifiers).

```js
const handleFetchErrors = createFetchErrorHandler({
    upstreamSystemCode: 'next-navigation-api'
});
```

When this is set, any errors thrown by `handleFetchErrors` will have a [`relatesToSystems` property](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/errors#operationalerrorrelatestosystems) which includes the given system code.

### DIFF
--- a/packages/fetch-error-handler/lib/create-handler.js
+++ b/packages/fetch-error-handler/lib/create-handler.js
@@ -72,7 +72,11 @@ function createFetchErrorHandler(options = {}) {
 				try {
 					const url = new URL(response.url);
 					responseHostName = url.hostname;
-				} catch (_) {}
+				} catch (_) {
+					// We ignore this error because having a valid URL isn't essential – it
+					// just helps debug if we do have one. If someone's using a weird non-standard
+					// `fetch` implementation or mocking then this error could be fired
+				}
 			}
 
 			// Some common error options which we'll include in any that are thrown
--- a/packages/fetch-error-handler/lib/create-handler.js
+++ b/packages/fetch-error-handler/lib/create-handler.js
@@ -1,0 +1,141 @@
+const {
+	HttpError,
+	UpstreamServiceError
+} = require('@dotcom-reliability-kit/errors');
+
+/**
+ * @typedef {object} ErrorHandlerOptions
+ * @property {string} [upstreamSystemCode]
+ *     The system code of the upstream system that the `fetch` makes a request to.
+ */
+
+/**
+ * @typedef {object} FetchResponse
+ * @property {boolean} ok
+ *     Whether the fetch was successful.
+ * @property {number} status
+ *     The response HTTP status code.
+ * @property {string} url
+ *     The URL of the response.
+ */
+
+/* eslint-disable jsdoc/valid-types */
+// The ESLint JSDoc plugin we're using marks this as invalid despite it being fine.
+// This is covered by the following GitHub issues. If these are closed then this should
+// no longer be an issue:
+// https://github.com/gajus/eslint-plugin-jsdoc/issues/145
+// https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/issues/96
+// https://github.com/jsdoctypeparser/jsdoctypeparser/issues/50
+/**
+ * @typedef {<Response extends FetchResponse>(fetchPromise: Response | Promise<Response>) => Promise<Response>} FetchErrorHandler
+ */
+/* eslint-enable jsdoc/valid-types */
+
+/**
+ * Create an fetch error handler function.
+ *
+ * @param {ErrorHandlerOptions} [options]
+ *     Configuration options for the handler.
+ * @returns {FetchErrorHandler}
+ *     Returns an error handler for use with a `fetch` function.
+ */
+function createFetchErrorHandler(options = {}) {
+	const { upstreamSystemCode } = options;
+	const relatesToSystems =
+		typeof upstreamSystemCode === 'string' ? [upstreamSystemCode] : [];
+
+	return async function handleFetchErrors(input) {
+		let response = input;
+
+		// If input is a promise, resolve it
+		if (input instanceof Promise) {
+			response = await input;
+		}
+
+		// Check whether the value we were given is a valid response object
+		if (!isFetchResponse(response)) {
+			// This is not an operational error because the invalid
+			// input is highly likely to be a programmer error
+			throw Object.assign(
+				new TypeError(
+					'Fetch handler must be called with a `fetch` response object or a `fetch` promise'
+				),
+				{ code: 'FETCH_ERROR_HANDLER_INVALID_INPUT' }
+			);
+		}
+
+		// If the response isn't OK, we start throwing errors
+		if (!response.ok) {
+			// Parse the response URL so we can use the hostname in error messages
+			let responseHostName = 'unknown';
+			if (typeof response.url === 'string') {
+				try {
+					const url = new URL(response.url);
+					responseHostName = url.hostname;
+				} catch (_) {}
+			}
+
+			// Some common error options which we'll include in any that are thrown
+			const baseErrorOptions = {
+				message: `The upstream service at "${responseHostName}" responded with a ${response.status} status`,
+				relatesToSystems,
+				upstreamUrl: response.url,
+				upstreamStatusCode: response.status
+			};
+
+			// If the back end responds with a `4xx` error then it normally indicates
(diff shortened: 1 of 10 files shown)


packages/fetch-error-handler/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@dotcom-reliability-kit/fetch-error-handler",
+  "version": "0.0.0",
```
CyntiBinti, line 3: suggestion: would the version need to change to v0.1.0 as it's experimental? or is this set to v0.0.0 due to release-please?

packages/fetch-error-handler/lib/create-handler.js
```diff
@@ -0,0 +1,141 @@
...
+// https://github.com/gajus/eslint-plugin-jsdoc/issues/145
+// https://github.com/jsdoc-type-pratt-parser/jsdoc-type-pratt-parser/issues/96
+// https://github.com/jsdoctypeparser/jsdoctypeparser/issues/50
+/**
+ * @typedef {<Response extends FetchResponse>(fetchPromise: Response | Promise<Response>) => Promise<Response>} FetchErrorHandler
+ */
+/* eslint-enable jsdoc/valid-types */
+
+/**
+ * Create an fetch error handler function.
```
CyntiBinti, line 35: nitpick: small typo from `an` to `a`

packages/fetch-error-handler/lib/create-handler.js
```diff
@@ -0,0 +1,141 @@
...
+
+		// If the response isn't OK, we start throwing errors
+		if (!response.ok) {
+			// Parse the response URL so we can use the hostname in error messages
+			let responseHostName = 'unknown';
+			if (typeof response.url === 'string') {
+				try {
+					const url = new URL(response.url);
+					responseHostName = url.hostname;
+				} catch (_) {}
```
CyntiBinti, line 75: curious: my understanding is that if we want to communicate to other devs that we've intentionally put a 'dummy' placeholder, then we use an underscore. Is this why the underscore is passed into the catch parameter? and if so, why do we have the catch and not just take the code out of the try/catch block? 🙇🏽

CyntiBinti (approved):
This looks super comprehensive and easy to read! - nice one 🎉

packages/fetch-error-handler/package.json
rowanmanning: The version will get bumped to `v0.1.0` by Release Please once this gets merged into `main` 🙂 setting it to `0.1.0` in this PR would cause Release Please to tag up `0.2.0`.

packages/fetch-error-handler/lib/create-handler.js
rowanmanning: Nice spot, thanks! I've fixed

packages/fetch-error-handler/lib/create-handler.js
CyntiBinti: Thank you as always 😸  lovely explanation and this makes sense. Yeah I think the comment above would be helpful to help clarify

rowanmanning:
I'm gonna merge this as it's low-risk to release a v0.1.0. We'll get it tested in some Customer Products repos at some point soon and report back any changes we decide to make.

PR #688, opened 2023-08-29
fix: stop requiring bound log methods

rowanmanning:
This caused a bug in next-subscribe who are providing a custom logger to Reliability Kit which does not have methods bound to `this`. Because we _can_ fix this in Reliability Kit I think we should - ideally we want as few code changes as possible when teams migrate.


CyntiBinti (approved):
🙌🏽  hopefully no more hourly alerts 😅

PR #694, opened 2023-09-04
docs: add a note on testing with Jest

### LINKED ISSUE #693: Add documentation on how to mock the logger with Jest
We need more documentation on how to mock the Reliability Kit logger with Jest.

The Content Discovery team had an annoying issue with the logger keeping worker handles open. I wasn't able to replicate anywhere _except_ the home page, I don't want to lose the knowledge gained as part of the investigation.

## Ideal solution

Add a section in the logger README about how to mock in Jest. [There's a long thread here which outlines the issue](https://financialtimes.slack.com/archives/C04UDE94K4H/p1689712510698239).

## Alternatives

I don't think there are alternatives, doing nothing will guarantee that we burn lots of time debugging the same issue in the future.

rowanmanning:
We found this issue in the FT.com home page when migrating to the Reliability Kit logger. The fix is to manually mock the logger when you're running tests using Jest. This documentation gives us something to point to if someone else encounters this issue.

Resolves #693.

### DIFF
--- a/packages/logger/README.md
+++ b/packages/logger/README.md
@@ -29,6 +29,7 @@ A simple and fast logger based on [Pino](https://getpino.io/), with FT preferenc
       * [Order of precedence](#order-of-precedence)
     * [Local development usage](#local-development-usage)
     * [Production usage](#production-usage)
+    * [Testing](#testing)
     * [Compatibility](#compatibility)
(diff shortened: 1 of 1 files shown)


metart43 (approved):
Nice. Thanks for adding this!

CyntiBinti (approved):

Issue #654, opened 2023-07-20
Add `processType` to the error logger


rowanmanning:
Add some more useful information (mainly Lambda name) to the error logger. Opening this on behalf of @huxley-ft who would like to log lambda name in crashes within The Edit.

## What problem does this feature solve?

For Serverless apps we don't currently log the name of the lambda that is erroring. In Splunk this is fine because the lambda name will appear as a `source` property but, if a team is trying to create alarms in AWS based on the actual CloudWatch logs, then they have no data to differentiate logs between different Lambda functions.

## Ideal solution

We already have a property under `appInfo` called [`processType`](https://github.com/Financial-Times/dotcom-reliability-kit/blob/a68e18f860b4c3d7579fbc0e84996d8e085cdecf/packages/app-info/lib/index.js#L61-L64). On Heroku this will be set to the dyno type (e.g. `web`), on Lambda it will be set to the value of the `AWS_LAMBDA_FUNCTION_NAME` environment variable.

I think we need to expose this data in the methods from the [log-error module](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/log-error#readme). We should be able to add `processType` to the [app object here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/a68e18f860b4c3d7579fbc0e84996d8e085cdecf/packages/log-error/lib/index.js#L59-L65).

## Alternatives

I don't think there's a neat alternative. We could possibly change this to be a different property depending on whether the app is running on Heroku or AWS, but I prefer the log structure to be as generic as possible.

rowanmanning:
Added in #708

PR #708, opened 2023-09-12
feat: add processType property to error logger

Shamarni96:
#654 
adds processType property to error logger to add extra error information when a lambda fails. 
ran `npm run tests` and all tests passed 

[Jira](https://financialtimes.atlassian.net/browse/CPREL-722)

### DIFF
--- a/packages/log-error/lib/index.js
+++ b/packages/log-error/lib/index.js
@@ -63,7 +63,8 @@ function logError({
 			name: appInfo.systemCode,
 			nodeVersion: process.versions.node,
 			region: appInfo.region,
-			releaseDate: appInfo.releaseDate
+			releaseDate: appInfo.releaseDate,
+			processType: appInfo.processType
 		}
 	};
 	if (request) {
--- a/packages/log-error/test/unit/lib/index.spec.js
+++ b/packages/log-error/test/unit/lib/index.spec.js
@@ -30,6 +30,7 @@ describe('@dotcom-reliability-kit/log-error', () => {
 		appInfo.region = 'mock-region';
 		appInfo.releaseDate = 'mock-release-date';
 		appInfo.systemCode = 'mock-system-code';
+		appInfo.processType = 'mock-process-type';
 	});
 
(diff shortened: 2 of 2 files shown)


rowanmanning (commented):
This is looking good, but although the tests are passing we're not ensuring that the `processType` continues to work and that it's taken from `appInfo`. I think we should also update [the test](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/packages/log-error/test/unit/lib/index.spec.js) to make sure that all references to the `app` object also contain mock data for the process type.

Let me know if you want to chat through this, but a starting point would be to find all references to `releaseDate` in the test file and ensure that we have an equivalent for `processType`. The mocking of the actual values is [here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/packages/log-error/test/unit/lib/index.spec.js#L32).

rowanmanning (commented):
This is looking great 🎉 thanks for sorting the tests! There's still a small issue with the commits, one is a `Merge` commit which we don't allow in this repo and the others are still `chore:` level rather than `feat:` level.

Have you done much rebasing before? I think squashing all three commits into a single commit named as follows would be best.

```
feat: add processType property to error logger
```

Then a force push to this branch. If you're not confident with that I'm happy to help out.

CyntiBinti:
> This is looking great 🎉 thanks for sorting the tests! There's still a small issue with the commits, one is a `Merge` commit which we don't allow in this repo and the others are still `chore:` level rather than `feat:` level.
> 
> Have you done much rebasing before? I think squashing all three commits into a single commit named as follows would be best.
> 
> ```
> feat: add processType property to error logger
> ```
> 
> Then a force push to this branch. If you're not confident with that I'm happy to help out.

@Shamarni96 I've found [this post](https://www.freecodecamp.org/news/git-squash-commits/) in the past super helpful with interactive rebasing and squashing commits -- happy to pair also on this!

CyntiBinti (approved):
Yay 🎉  nice one with the rebase!

PR #716, opened 2023-09-18
feat: handle common errors thrown by fetch

rowanmanning:
This adds in handling for most of the common errors that we see when making `fetch` requests. There's a lot of extra logic in here but I think it's worthwhile to augment these errors and throw ones with correct status codes that our engineers can immediately understand.

Due to the fact that we use a variety of different `fetch` implementations across the FT, this really isn't doable at the Splunk level: we can't reliably say "here's a search which shows every fetch request that has timed out".

Because of the amount of logic and the fact we're relying on error names and codes, this comes with a suite of end-to-end tests that allow us to verify that the following `fetch` implementations throw the expected errors. This way, if something changes, we'll see it immediately in a failing dependabot PR:

  - Node.js 18+ native fetch (powered by Undici)
  - Manual [Undici](https://github.com/nodejs/undici) installs
  - [node-fetch](https://github.com/node-fetch/node-fetch) v1.x
  - [node-fetch](https://github.com/node-fetch/node-fetch) v2.x

I'm unable to test using node-fetch v3 because it is written in native ES Modules which Jest has very patchy support for. I think this is acceptable as, because of the native ESM, most of our code uses node-fetch v2 and below. I think over time we're more likely to move towards native `fetch` than spend time upgrading our systems to use node-fetch v3.

### DIFF
--- a/packages/fetch-error-handler/README.md
+++ b/packages/fetch-error-handler/README.md
@@ -7,9 +7,15 @@ Properly handle fetch errors and avoid a lot of boilerplate in your app. This mo
 > This package is in beta and hasn't been tested extensively in production yet. Feel free to use, and any feedback is greatly appreciated.
 
   * [Usage](#usage)
+    * [Wrap the fetch function](#wrap-the-fetch-function)
+    * [Handle errors with `.then`](#handle-errors-with-then)
+    * [Handle the response object](#handle-the-response-object)
     * [Errors thrown](#errors-thrown)
       * [Client errors](#client-errors)
       * [Server errors](#server-errors)
+      * [DNS errors](#dns-errors)
+      * [Abort and timeout errors](#abort-and-timeout-errors)
+      * [Socket errors](#socket-errors)
       * [Unknown errors](#unknown-errors)
     * [Creating your own handler](#creating-your-own-handler)
     * [`createFetchErrorHandler` configuration options](#createfetcherrorhandler-configuration-options)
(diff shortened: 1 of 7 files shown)


packages/fetch-error-handler/README.md
```diff
@@ -36,26 +42,41 @@ const { handleFetchErrors } = require('@dotcom-reliability-kit/fetch-error-handl
...
-There are many ways to use it, as long as it is `await`ed and is called with either a `Response` object or a `Promise` that resolves to a `Response`. The function is asynchronous and resolves with the `Response` that it was called with:
+There are several ways to use it, as long as it is `await`ed and is called with either a `Response` object or a `Promise` that resolves to a `Response`.
 
-```js
-// Pass the function into the `then` function of a `fetch`.
-// Note: this must be `then` and not `catch`
-const response = await fetch('https://httpbin.org/status/500').then(handleFetchErrors);
+Some of the options below result in more errors being caught, you can weigh this up when implementing in your own code.
+
+In all of the APIs below, if the reponse `ok` property is `false`, i.e. when the status code is `400` or greater, then errors will be thrown.
```
CyntiBinti, line 49: suggestion: small typo on response

```suggestion
In all of the APIs below, if the response `ok` property is `false`, i.e. when the status code is `400` or greater, then errors will be thrown.
```

packages/fetch-error-handler/test/unit/lib/create-handler.spec.js
```diff
@@ -168,6 +169,175 @@ describe('@dotcom-reliability-kit/fetch-error-handler', () => {
...
+						expect(error.code).toStrictEqual('FETCH_DNS_LOOKUP_ERROR');
+						expect(error.message).toStrictEqual(
+							'Cound not resolve DNS entry for host mockhost'
+						);
+						expect(error.cause).toStrictEqual(mockError);
+					}
+				});
+			});
+
+			describe('when the promise rejects with a abort error', () => {
```
CyntiBinti, line 241: nitpick: small typo

```suggestion
			describe('when the promise rejects with an abort error', () => {
```

CyntiBinti (approved):
Looking great! 🙌🏽

PR #736, opened 2023-10-10
fix: trim long request URLs

rowanmanning:
When the request URL is particularly long, Splunk events break because they're concatenated if they exceed a certain length. This causes the JSON to be invalid and so we don't parse out fields.

We probably shouldn't blindly trust the URL as it's user input. We now truncate request URLs at 200 characters and add an indicator so that it's clear in the logs when the URL was too long.

### DIFF
--- a/packages/serialize-request/test/unit/lib/index.spec.js
+++ b/packages/serialize-request/test/unit/lib/index.spec.js
@@ -197,6 +197,19 @@ describe('@dotcom-reliability-kit/serialize-request', () => {
 			});
 		});
 
+		describe('when `request.url` is longer than 200 characters', () => {
+			beforeEach(() => {
+				request.url =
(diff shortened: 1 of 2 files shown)


packages/serialize-request/test/unit/lib/index.spec.js
```diff
@@ -197,6 +197,19 @@ describe('@dotcom-reliability-kit/serialize-request', () => {
 			});
 		});
 
+		describe('when `request.url` is longer than 200 characters', () => {
+			beforeEach(() => {
+				request.url =
+					'/url-that-is-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-long';
```
CyntiBinti, line 203: 😸 robust!

CyntiBinti (approved):

alexmuller (approved):

PR #765, opened 2023-10-30
feat: add cloudProvider as exported property for appInfo

CyntiBinti:
Adds a cloudProvider property which is set to either heroku or aws depending on where we think the system is hosted. To figure this out we’ll have to find some environment variables which we have high confidence are defined by only Heroku or AWS/Lambda.

We’re going to use this in the OpenTelemetry tracing.

[JIRA](https://financialtimes.atlassian.net/browse/CPREL-834)

### DIFF
--- a/packages/app-info/lib/index.js
+++ b/packages/app-info/lib/index.js
@@ -134,8 +134,11 @@ module.exports = {
 	 * @type {string | null}
 	 */
 	cloudProvider:
-		(processType === process.env.AWS_LAMBDA_FUNCTION_NAME ? 'aws' : 'heroku') ||
-		null
+		(processType === process.env.AWS_LAMBDA_FUNCTION_NAME
+			? 'aws'
+			: processType === process.env.DYNO
+				? 'heroku'
+				: null) || null
 };
 
 // @ts-ignore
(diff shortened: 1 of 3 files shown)


packages/app-info/lib/index.js
```diff
@@ -125,7 +125,20 @@ module.exports = {
...
+	 * The cloud provider type.
+	 *
+	 * @readonly
+	 * @type {string | null}
+	 */
+	cloudProvider:
+		(processType === process.env.AWS_LAMBDA_FUNCTION_NAME
+			? 'aws'
+			: processType === process.env.DYNO
+				? 'heroku'
```
CyntiBinti, line 140: eslint is not passing on line 140 and 141 due to it expecting 4 tabs but got 3, but when I change it to 4 tabs then prettier complains 🥲 . Got around this by passing `--no-verify` in the terminal but will need to fix for CircleCI to pass I'd imagine

CyntiBinti: Changed code completely

alexmuller (approved):
Looks great! Rowan may have an idea for another Heroku environment variable we can use but for now I think we should go with this 👍 👍

PR #783, opened 2023-11-13
feat: add some basic app info to error pages

rowanmanning:
This is a first step towards serving nicer error pages in production. We want to show application information so that a non-engineer would be able to tell which system an error is served by. This doesn't make error pages available in production yet, that'd be a later step if we decide we want to add it.

In the meantime this makes the error page more useful.

It looks like this:

Screenshot 2023-11-13 at 14 06 40


CyntiBinti (approved):
Yay for richer information  🙌🏽 🚀

PR #785, opened 2023-11-14
fix: handle more error rendering edge-cases

rowanmanning:
This handles a couple of edge-cases we were missing that we _do_ handle in n-express. These are required before we can consider replacing the n-express error handling with Reliability Kit. We've done the following:

  * Ensure that we don't try to render the error page if the response headers have already been sent. This is recommended in the Express documentation

  * Ensure that the HTTP status code we send is between 400 and 599. This prevents poor error handling in apps from accidentally sending false positives by responding with 200 statuses for errors.


CyntiBinti (approved):
LGTM! 👌🏽

PR #787, opened 2023-11-14
feat: add a fingerprint to serialized errors

rowanmanning:
This helps us uniquely identify errors and gives us another data point to group errors by without resorting to very complicated Splunk logic. We're intending on using this fingerprint to help Customer Care and Operations more quickly direct us to the root cause of a problem.

You will also be able to more easily see the common errors that your application throws if you're using the Reliability Kit logging middleware.


alexmuller (approved):
This looks great! Super interested to see what happens with this in Splunk and figure out how we can iterate on it to make errors easy to track down.

PR #802, opened 2023-11-21
fix: switch to using streams for prettification

rowanmanning:
We encountered an issue in pino-pretty where, if the `--require` argument is used, the required file will be loaded twice. This can cause some weird and unexpected issues. We raised a bug with pino-pretty here: https://github.com/pinojs/pino-pretty/issues/478

Switching to using pino-pretty as a stream resolves this issue.


CyntiBinti:
Discovered from this PR where we're exploring open Telemetry and AWS X-ray: #https://github.com/Financial-Times/dotcom-debug-user/pull/196

CyntiBinti (approved):
Thaaanks!

PR #807, opened 2023-11-23
feat: add useIsoTimeFormat option to format log timestamps

alexmuller:
Simeon reported in the cp-reliability Slack channel that they would prefer to have more human readable timestamps in logs. Pino supports this with the stdTimeFunctions object which can be set on the withTimestamps option (that we're using already).

This commit exposes the Pino configuration with a new Reliability Kit logger option called useIsoTimeFormat.

We anticipate that in the future we will probably switch this to be the default (which will be a breaking change) and potentially remove the option completely so that timestamps are always logged.

Note that this timestamp option is on the log body. We do already have the log ingestion datetime recorded, even when withTimestamps is false.


rowanmanning (approved):
Nice 👏

SimeonAleksandrovFT (commented):
Thanks a lot!

PR #821, opened 2023-12-05
Feature/add application

awar7118:
In the direction of using openTelemetry, this adds the `herokuDynoId` and `herokuAppId` properties to the appInfo object.

These properties come as part of the dyno metadata.

Read more here --> [Link](https://devcenter.heroku.com/articles/dyno-metadata)


packages/app-info/README.md
```diff
@@ -11,7 +11,8 @@ A utility to get application information (e.g. the system code) in a consistent
     * [`appInfo.releaseVersion`](#appinforeleaseversion)
     * [`appInfo.systemCode`](#appinfosystemcode)
     * [`appInfo.processType`](#appinfoprocesstype)
-    * [`appInfo.cloudProvider](#appinfocloudprovider)
```
rowanmanning, line 14: **issue:** looks like we deleted the cloud provider docs link by accident.

packages/app-info/README.md
```diff
@@ -84,6 +85,18 @@ Get the type of cloud provider which is set to either Heroku or AWS depending on
 
 This is derived from some environment variables which we have high confidence are defined by only Heroku or AWS/Lambda
 
+### `appInfo.herokuAppId`
+
+Get the HEROKU_APP_ID which is the unique identifier for the application.e.g "9daa2797-e49b-4624-932f-ec3f9688e3da"
```
rowanmanning, line 90: **suggestion:** could we backtick the environment variable names so they're formatted as code blocks? Also maybe prefix with `process.env.` to match the documentation for the other properties. Same for the `appInfo.herokuDynoId` docs below.

packages/app-info/lib/index.js
```diff
@@ -44,7 +44,7 @@ function normalizePackageName(name) {
 }
 
 /**
- * Extract the proceess type from a Heroku dyno name.
+ * Extract the process type from a Heroku dyno name.
```
rowanmanning, line 47: **praise:** nice spot 😄

rowanmanning (commented):
A few small changes needed in the docs but otherwise this is looking great 👏

In order for us to retain our conventional commit, you'll need to add the changes to it with `git commit --amend`.

packages/app-info/README.md
awar7118: oops nice spot. Have re-added it now

PR #840, opened 2023-12-22
chore!: drop support for Node.js 16 and npm 7

### LINKED ISSUE #806: Drop support for Node.js 16
Officially drop support for Node.js 16 and stop running tests against it.

Node.js 16 is end-of-life, we can simplify and speed up our testing and discourage use of an unsupported version of Node.js.

## Ideal solution

We remove Node.js 16 support and release new major versions of every package.

## Alternatives

There are no alternatives.

rowanmanning:
This can't be merged until January 2024 when we're dropping Node.js 16 support from packages. Just getting it ready to save us a job in the new year.

This resolves #806.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,21 +157,21 @@ workflows:
           name: build-v<< matrix.node-version >>
           matrix:
             parameters:
-              node-version: [ "20.0", "18.16", "16.20" ]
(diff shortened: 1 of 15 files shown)


awar7118 (approved):
Looks good, ran `npm install` with `node 16` and `node 18` and have seen the support difference

PR #851, opened 2024-01-05
docs: add migration guides for all packages

rowanmanning:
There's a small maintenance overhead to adding these but I think they're far more clear than a changelog for major breaking versions. We need to include a lot of extra context and information and I think these guides will reduce questions from other teams.


CyntiBinti (approved):
praise: really appreciate your documentation being consistently thorough and thoughtful - - thank you! 🙌🏽

PR #863, opened 2024-01-08
docs: replace node 16 references with node 18

awar7118:
To avoid confusion, removed references to Node 16 in docs, now that node16 support bas been dropped by Dotcom-Reliability-kit-> #840


rowanmanning (approved):
Nice 🙂 good spot

PR #864, opened 2024-01-08
fix: add official support for npm 10

rowanmanning:
We missed this when it was released.


joelcarr (approved):

alexmuller (approved):
👍 Will need adding to the new opentelemetry package as I think you mentioned in Slack.

PR #883, opened 2024-01-16
docs: add a README for the opentelemetry package

rowanmanning:
This adds full documentation for the opentelemetry package.


packages/opentelemetry/README.md
```diff
@@ -4,4 +4,181 @@
...
+* [Contributing](#contributing)
+* [License](#license)
+
+
+## Usage
+
+Install `@dotcom-reliability-kit/opentelemetry` as a dependency:
+
+```bash
+npm install --save @dotcom-reliability-kit/opentelemetry
```
CyntiBinti, line 26: non-blocking: we only want it as a dev dependency? We don't want it to have access to it when running in production?

CyntiBinti (approved):
Beautiful 👨🏽‍🍳

packages/opentelemetry/README.md
rowanmanning: Nope we want it as a prod dependency which `--save` does 🙂

PR #895, opened 2024-01-19
docs: add running instructions

rowanmanning:
This outlines how to use OpenTelemetry in production and local development, and what config changes you might need. I think this is worth documenting as I'm sure we'll get questions relating to this.

I'm deliberately not outlining how we do this in production in Customer Products:

1. This can be used by other teams
2. This will change once we have a central store
3. It's a public repo, we don't want to share our tracing setup here

### DIFF
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -3,13 +3,21 @@
 
 An [OpenTelemetry](https://opentelemetry.io/docs/what-is-opentelemetry/) client that's preconfigured for drop-in use in FT apps. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
 
+> [!TIP]<br />
+> OpenTelemetry is an open source observability framework that supports sending metrics, traces, and logs to a large variety of backends via a shared protocol. We try to abstract some of these concepts away with this module, but [understanding OpenTelemetry will help you get it set up](https://opentelemetry.io/docs/what-is-opentelemetry/).
+
 > [!WARNING]<br />
 > This package is in beta and hasn't been tested extensively in production yet. Feel free to use, and any feedback is greatly appreciated.
 
 * [Usage](#usage)
-  * [Automated setup with `--require`](#automated-setup-with---require)
-  * [Automated setup with `require()`](#automated-setup-with-require)
-  * [Manual setup](#manual-setup)
+  * [Setup](#setup)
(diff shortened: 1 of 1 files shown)


CyntiBinti (dismissed):

CyntiBinti:
Curious: For repo's like debug-user that don't import the CPREL OTel function into it's `./lib/server/app.js,` but instead `--require` via it's Procfile, how would they customise the arguments passed in to run locally?

rowanmanning:
For those you'll need to use the environment variables :) this is what I ran:

```
OPENTELEMETRY_TRACING_ENDPOINT=https://localhost:4318/v1/traces OPENTELEMETRY_TRACING_SAMPLE_PERCENTAGE=100 npm start
```

CyntiBinti:
> For those you'll need to use the environment variables :) this is what I ran:
> 
> ```
> OPENTELEMETRY_TRACING_ENDPOINT=https://localhost:4318/v1/traces OPENTELEMETRY_TRACING_SAMPLE_PERCENTAGE=100 npm start
> ```

suggestion: would it be worth adding this as an example command (before line 174) to the README?

packages/opentelemetry/README.md
```diff
@@ -135,6 +145,35 @@ setupOpenTelemetry({ /* ... */ });
...
+[Jaeger maintains a useful guide for this](https://www.jaegertracing.io/docs/1.53/getting-started/#all-in-one).
+
+#### Sending traces to your local backend
+
+Once your backend is running you'll need to make some configuration changes.
+
+You'll need to set the [tracing endpoint](#optionstracingendpoint) to use Jaeger's tracing endpoint on port `4318` ([OTLP/HTTP](https://opentelemetry.io/docs/specs/otlp/#otlphttp)). E.g. `http://localhost:4318/v1/traces`.
+
+You'll also need to disable sampling by [configuring it](#optionstracingsamplepercentage) to `100`.
+
```
CyntiBinti, line 173: suggestion: would it be worth adding this as an example command (before line 174) to the README?

```suggestion
OPENTELEMETRY_TRACING_ENDPOINT=http://localhost:4318/v1/traces OPENTELEMETRY_TRACING_SAMPLE_PERCENTAGE=100 npm start
```

CyntiBinti (approved):

packages/opentelemetry/README.md
rowanmanning: Good shout 👍 I've updated this

PR #898, opened 2024-01-22
feat: ignore commonly polled URLs

rowanmanning:
This excludes commonly polled URLs from the tracing. By doing this at the instrumentation level we shouldn't impact the percentage of traces that are sampled and we'll see a lot less noise.


alexmuller (approved):
Looks right to me!

PR #899, opened 2024-01-22
docs!: mark as stable

CyntiBinti:
We've tested out this beta package in next-preflight, dotcom-debug-user, and also reliability-dashboards, and feel happy to mark it as stable for other systems to install now in production. This completes CPREL-956

Release-as: 1.0.0


rowanmanning (approved):
Nice 👏

PR #925, opened 2024-01-31
fix: handle invalid URLs in the request filter

rowanmanning:
Request URLs can still be invalid and make it all the way into our filter. One way you can cause this error to occur is by not setting a `host` header. We don't want errors to be logged for this so I'm now logging a warning with details about the URL that failed to be parsed.


CyntiBinti (approved):
Noice 🔥

Issue #303, opened 2022-11-03
Write tests to increase confidence in different module systems


rowanmanning:
We need a test suite that increases our confidence that Reliability Kit works regardless of which module system our consuming apps use.

## What problem does this feature solve?

We've had a bit of a nightmare with supporting different module systems. Many of our issues have been due to type hinting being wrong or modules not being importable as we expected based on the way an app is set up. I think we need to increase our confidence that Reliability Kit will work across the majority of our Node.js projects. Ideally we should do this without introducing _too_ much overhead.

## Ideal solution

Write a [custom eslint plugin](https://eslint.org/docs/latest/extend/plugins) that would check our reliability-kit `.js ` files for the presence of a `module.exports` and if present, it should also have a corresponding `module.exports.default`. If the `module.exports.default` is not present then eslint will error and the commit will fail as part of the husky process.

## Alternatives

We should write some end-to-end tests, either at the package level or in the root of the project, which imports and checks each of the expected public exports. We need each module to work in at least the following:

  * In a CommonJS app (non-compiled)
  * In an [ESM](https://nodejs.org/api/esm.html) app (non-compiled)
  * In a TypeScript app that outputs CommonJS ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `false`)
  * In a TypeScript app that outputs ESM ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `true`)
  * In a TypeScript app that outputs CommonJS ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `false`)
  * In a TypeScript app that outputs ESM ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `true`)
  * In a CommonJS app which uses [Sucrase](https://sucrase.io/)
  * In a CommonJS app which is compiled with [Babel](https://babeljs.io/)

We could also do nothing, but I'm finding that my development is slowed down because I have to manually test in a few different module systems locally to be sure I'm not introducing a breaking change.

rowanmanning:
I'm unsure if this is as high a priority right now as we haven't had any issues despite adoption across the group increasing. Gonna move back to the inbox for now and we can pick up later if the logger roll-out surfaces any issues.

rowanmanning:
I'm going to consider this as done, we have tests in place and we'll add new ones for the other modules once we switch to manual type declarations.

PR #927, opened 2024-02-01
Fix the types (part 1)

rowanmanning:
Sorry for the size of this PR, it was really difficult to split but we're now at the point where we can introduce manual type definitions on a per-module basis. This will allow us to fix specific type issues that people are having when they come up.

I've added some tests to verify that we have a problem. I've tried to replicate a few different module formats that we use at the FT to make sure we can get this working with all of them:

  * In a CommonJS app (non-compiled)
  * In an [ESM](https://nodejs.org/api/esm.html) app (non-compiled)
  * In a TypeScript app that outputs CommonJS ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `false`)
  * In a TypeScript app that outputs ESM ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `true`)
  * In a TypeScript app that outputs CommonJS ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `false`)
  * In a TypeScript app that outputs ESM ([`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop) set to `true`)
  * In a CommonJS app which uses [Sucrase](https://sucrase.io/)

This is the first step towards having more confidence in our support of different module systems. This forms part of #303.

Various module formats don't work, e.g. the logger types don't work well for TypeScript projects that output CommonJS. The serialize-request types don't work for most ESM. I've added `@ts-ignore` statements to make the tests pass and document the places where we have inconsistency so that we can start to address specific type issues.

The last part of this PR is taking a currently working set of types (app-info) and reworking them so that the type definitions are manually created and exported. In other modules this will reduce the amount of JSDoc type declarations which is great. It also gives us better hinting in VSCode because we're naming the modules.

---

Relates-to: CPREL-972


packages/app-info/package.json
```diff
@@ -14,5 +14,6 @@
     "node": "18.x || 20.x",
     "npm": "8.x || 9.x || 10.x"
   },
-  "main": "lib"
+  "main": "lib/index.js",
+  "types": "types/index.d.ts"
```
alexmuller, line 18: Should this change also go in `scripts/create-package.js`?

rowanmanning: Yep 👍

CyntiBinti (approved):
Praise: Definitely not an easy PR to have written especially going through the journey of getting it to this point - nice job on persevering with it 🙌🏽

alexmuller (approved):
Thanks!

PR #942, opened 2024-02-20
fix: add type declarations for serialize-error

rowanmanning:
This adds manual type declarations for serialize-error which should make it work in most projects.

See-also: [CPREL-985](https://financialtimes.atlassian.net/browse/CPREL-985)

[CPREL-985]: https://financialtimes.atlassian.net/browse/CPREL-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ packages/logger/**/*.d.*
 packages/middleware-log-errors/**/*.d.*
 packages/middleware-render-error-info/**/*.d.*
 packages/opentelemetry/**/*.d.*
-packages/serialize-error/**/*.d.*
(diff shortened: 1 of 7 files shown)


rowanmanning:
Yep there are already tests for this package but they were passing so there was no need to change them 👍 I'm mostly moving to manual types for consistency with this one and because it reduces the amount of code we maintain

CyntiBinti (approved):

PR #944, opened 2024-02-20
fix: add type declarations for eslint-config

rowanmanning:


CyntiBinti:
question: are we doing type checking tests for this package? like with previous packages

CyntiBinti (approved):

PR #965, opened 2024-03-01
DEMO Do not merge: add metric reader with oTel exporter

ytcleon:
## Description

This PR is a part of the SPIKE ticket to investigate the filtering and aggregation of Otel traces and metrics. It is for demonstration purpose only.

## Changes

When a metric endpoint (e.g. http://localhost:4318/v1/metrics) is provided in `OPENTELEMETRY_METRICS_ENDPOINT`, it adds a metric reader with oTel exporter to the oTel SDK, so the app can send metrics to the endpoint.

## Note

Although it is planned that Amazon Managed Service for Prometheus will be used as the backend of metric storage, it would be better to output the metrics in oTel format and convert the format to Prometheus in oTel collector, rather than directly using Prometheus exporter in the metric reader.

### DIFF
--- a/packages/opentelemetry/lib/index.js
+++ b/packages/opentelemetry/lib/index.js
@@ -14,6 +14,11 @@ const {
 } = require('@opentelemetry/exporter-trace-otlp-proto');
 const traceExporterPackageJson = require('@opentelemetry/exporter-trace-otlp-proto/package.json');
 const {
+	OTLPMetricExporter
+} = require('@opentelemetry/exporter-metrics-otlp-proto');
+const matricExporterPackageJson = require('@opentelemetry/exporter-metrics-otlp-proto/package.json');
+const { PeriodicExportingMetricReader } = require('@opentelemetry/sdk-metrics');
+const {
 	NoopSpanProcessor,
 	TraceIdRatioBasedSampler
 } = require('@opentelemetry/sdk-trace-base');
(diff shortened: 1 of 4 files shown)


packages/opentelemetry/lib/index.js
```diff
@@ -150,6 +164,39 @@ function setupOpenTelemetry({ authorizationHeader, tracing } = {}) {
...
+		if (authorizationHeader) {
+			headers.authorization = authorizationHeader;
+		}
+		const exporter = new OTLPMetricExporter({
+			url: metrics?.endpoint,
+			headers
+		});
+		openTelemetryConfig.metricReader = new PeriodicExportingMetricReader({
+			exporter,
+			exportIntervalMillis: 1000 // Set as 1000ms for development and demostration, need to set it higher for prod
```
sjparkinson, line 180: Leave this to the default value when you move forwards with this. Unlike Graphite we will crack down on teams shipping metrics more frequently than the default 60 seconds.

#### References

* https://github.com/open-telemetry/opentelemetry-js/blob/82b7526b028a34a23936016768f37df05effcd59/packages/sdk-metrics/src/export/PeriodicExportingMetricReader.ts#L97

packages/opentelemetry/lib/index.js
```diff
@@ -21,6 +26,7 @@ const logger = require('@dotcom-reliability-kit/logger');
 
 const USER_AGENT = `FTSystem/${appInfo.systemCode} (${packageJson.name}/${packageJson.version})`;
 const TRACING_USER_AGENT = `${USER_AGENT} (${traceExporterPackageJson.name}/${traceExporterPackageJson.version})`;
+const METRIC_USER_AGENT = `${USER_AGENT} (${matricExporterPackageJson.name}/${matricExporterPackageJson.version})`;
```
sjparkinson, line 29: Is the thinking here to distinguish this package from the default user-agent set by the OTel SDK?

If so it might be worth being aware of resource attributes, there's a few set by the SDK:

https://opentelemetry.io/docs/specs/semconv/resource/#telemetry-sdk

Maybe you want to utilise them, or something similar to keep track of what's in use? Very keen to discuss this if you want to do something different to the defaults, so we can get everyone on the same plan.

sjparkinson (commented):
Adding some FYI comments.

rowanmanning:
Hi @ytcleon, thanks for doing this, it's been useful as a learning exercise 🙂 I'm going to close this PR as we now have a finished implementation in #1089. Fingers crossed we'll ship it soon!

PR #977, opened 2024-03-21
fix: bump all OpenTelemetry packages

rowanmanning:
None of the individual dependency PRs pass the tests because they're not compatible with the other versions. This bumps everything at once.

There was also a deprecation in semantic resource attributes which was causing some type errors. I've fixed these.


CyntiBinti (approved):
Nice find with the deprecation - thank you! ✨

PR #981, opened 2024-03-22
Fix the changelogs

rowanmanning:
I've been wanting to do this for ages and Release Please have finally fixed the issue their end. When dependencies were updated _within_ the monorepo the changelogs didn't get updated correctly with a version heading. I've regenerated all the impacted changelog files based on the GitHub release data.

The bulk of the work was done with a quick script I wrote after pulling the release JSON. I double checked myself and think everything is correct 👍


CyntiBinti (approved):
Nice one! 🚀

PR #1010, opened 2024-04-23
fix: handle node-fetch memory leak

rowanmanning:
node-fetch has a memory leak if you don't read the body of a response. Within the library we don't ever know which fetch implementaton we're using so we've had to test for a property (`pipe`) that only exists on the node-fetch response body and not native fetch.

If we're throwing an error and we see that the response body has this method, then we pipe the body into a black hole stream. This means that the body is unusable if there was an error but we can document this limitation. I'd rather not mess with trying to read/parse data from the response body if possible as it adds a lot of complexity and I don't want to add _too_ many cases to deal with node-fetch when we'll be moving to native-fetch relatively soon.

We discussed some alternatives but I decided the best thing to do for now is to fix the memory leak and then we can deal with adding the response body to the thrown errors as a separate feature request.

**Note:** this PR has a bunch of whitespace changes because I had to reindent code to introduce another `try/catch`. It'll be easier to review with [whitespace changes ignored](https://github.com/Financial-Times/dotcom-reliability-kit/pull/1010/files?w=1).

See-Also: [CPREL-1047](https://financialtimes.atlassian.net/browse/CPREL-1047)

[CPREL-1047]: https://financialtimes.atlassian.net/browse/CPREL-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/fetch-error-handler/test/unit/lib/create-handler.spec.js
+++ b/packages/fetch-error-handler/test/unit/lib/create-handler.spec.js
@@ -1,6 +1,9 @@
-const createFetchErrorHandler = require('../../../lib/create-handler');
+jest.mock('node:stream');
 
-describe('@dotcom-reliability-kit/fetch-error-handler', () => {
+const createFetchErrorHandler = require('../../../lib/create-handler');
+const { Writable } = require('node:stream');
+
+describe('@dotcom-reliability-kit/fetch-error-handler/lib/create-handler', () => {
 	let fetchErrorHandler;
 
 	describe('createFetchErrorHandler()', () => {
@@ -20,7 +23,8 @@ describe('@dotcom-reliability-kit/fetch-error-handler', () => {
 				mockResponse = {
 					ok: true,
 					status: 200,
-					url: 'https://mock.com/example'
+					url: 'https://mock.com/example',
+					body: {}
 				};
 				resolvedValue = await fetchErrorHandler(mockResponse);
 			});
@@ -148,7 +152,8 @@ describe('@dotcom-reliability-kit/fetch-error-handler', () => {
 				mockResponse = {
 					ok: true,
 					status: 200,
-					url: 'https://mock.com/example'
+					url: 'https://mock.com/example',
+					body: {}
 				};
 				resolvedValue = await fetchErrorHandler(Promise.resolve(mockResponse));
 			});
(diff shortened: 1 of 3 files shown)


kavanagh:
I think this solution is good for when you don't care about the response body (which is fairly common at the FT as many APIs don't response with anything useful) and you're not using `createFetchErrorHandler`.

However some APIs do respond with useful information. We can filter what's contained in the body and attach to the error providing us with much richer info in logs. Sometime this data can be used to improve the quality/accuracy of the error message or code.

> An alternative would be to read the response body as text and attach it to the error object so that it can be read by an app if required / attempt to have it parsed as JSON.

This could end up holding onto an awful lot of rubbish and that presumably becomes `error.data.response` and will be serialised by the logger?

I think my ideal would be something like

```
createFetchErrorHandler({
  errorProperties: ({ headers, json, text }) {
    // here I can read data about the response after the body has bead read
    // and return data which is then used to decorate/modify the thrown error.
    return {
      code: 'FOO_BAR', // I can override the code
      message: 'Foo bar error', // or message
      data: {} // or capture extra data depending on what the content of the headers and body
    }
  }
})
```

kavanagh:
Another option is to change the API. Golang could be an inspiration here:

```
const [error, response] = await handleFetchErrors(
    fetch('https://httpbin.org/status/500')
);

if (error) {
 // do stuff to handle the error.
 // perhaps use it as the cause for another error
 // then throw or log the error
}
```

packages/fetch-error-handler/test/unit/lib/create-handler.spec.js
```diff
@@ -168,6 +173,34 @@ describe('@dotcom-reliability-kit/fetch-error-handler', () => {
...
+
+							// We make sure the _write function doesn't do anything with
+							// any of the data aside from calling `done`
+							const chunk = {}; // The test will error if we try to call any chunk methods, e.g. toString
+							const encoding = {}; // The test will error if we try to perform string operations on the encoding
+							const done = jest.fn();
+							stream._write(chunk, encoding, done);
+							expect(done).toHaveBeenCalledTimes(1);
+							expect(done).toHaveBeenCalledWith();
+						}
```
CyntiBinti, line 201: Is the `.toHaveBeenCalledWith()` meant to be empty?

rowanmanning: Yep 👍  because we expect `done` to be called with no arguments. Unfortunately there's not a nice Jesty way to do this where it's clearer what we're doing. An alternative would be this but I think it's possible less clear, thoughts?

```js
expect(done.mock.calls[0].length).toEqual(0);
```

Alternatively I could add a comment to clarify if you like?

CyntiBinti: Ah I see, ok this makes sense. Now explained, the original code is fine. I think a comment would suffice 😌

CyntiBinti (approved):

PR #1021, opened 2024-04-26
feat: add support for Node.js 22

rowanmanning:
I feel confident enough in our end-to-end tests to add support immediately. v22 will become an LTS release so it's worth testing against it as soon as possible.

### DIFF
--- a/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
+++ b/packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
@@ -7,11 +7,16 @@
  *     Returns the parsed JSON logs as an array of log objects.
  */
 function splitAndParseJsonLogs(logString) {
-	return logString
-		.trim()
-		.split(/\n+/g)
-		.filter((log) => log.trim())
-		.map((log) => JSON.parse(log));
+	return (
+		logString
+			.trim()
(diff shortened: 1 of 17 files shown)


packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
```diff
@@ -7,11 +7,16 @@
...
-		.filter((log) => log.trim())
-		.map((log) => JSON.parse(log));
+	return (
+		logString
+			.trim()
+			.split(/\n+/g)
+			.filter((log) => log.trim())
+			// Node.js sometimes logs deprecation warnings which are not valid JSON.
+			// We need to filter these out so we don't error in our JSON.parse map
+			.filter((log) => !log.toLowerCase().includes('deprecation'))
```
CyntiBinti, line 17: Nice addition here!

CyntiBinti (approved):
🚀

packages/logger/test/end-to-end/helpers/split-and-parse-json-logs.js
rowanmanning: Yep Node.js 22 deprecates some stuff so the tests were failing 😁

PR #1025, opened 2024-05-01
fix: make FetchResponse type compatible with a native Response object

apaleslimghost:


packages/fetch-error-handler/lib/create-handler.js
```diff
@@ -25,7 +25,7 @@ const { Writable } = require('node:stream');
  *     The response HTTP status code.
  * @property {string} url
  *     The URL of the response.
- * @property {FetchResponseBody} body
+ * @property {NodeFetchResponseBody | ReadableStream<Uint8Array> | null} body
```
apaleslimghost, line 28: `ReadableStream` here is the native web streams type, which comes from the DOM lib types. i'm not 100% certain how this will behave if the consumer doesn't have that lib.

apaleslimghost: actually this seems to work fine in cp-content-pipeline-schema

CyntiBinti (approved):

PR #1045, opened 2024-05-17
fix: add log-error as dependency of opentelemetry

alexmuller:
We use logRecoverableError() in the OpenTelemetry package.

Thanks @huxley-ft for spotting this!


rowanmanning (approved):
Oops! Thanks for fixing

PR #1054, opened 2024-05-23
chore: bump all OpenTelemetry packages at once

rowanmanning:
We see a lot of PRs for OpenTelemetry packages because many of them are technically breaking changes - e.g. bumping from 0.4.0 to 0.5.0. It's annoying and I have to do lots of extra work to keep these packages up to date. This should mean we only get one.


alexmuller (approved):
This is an excellent piece of work Rowan

PR #1066, opened 2024-05-31
Refactor the OpenTelemetry package

rowanmanning:
This refactors the package, splitting things into more testable modules ahead of adding in metrics support. I tried adding support already but the tests get unwieldy and it was difficult to reason about anything; I took a step back and did some refactoring which will make adding metrics a lot easier.

I tested this manually as well by linking it into a local project and testing that traces come through in Jaeger.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1900,11 +1900,11 @@
       }
     },
     "node_modules/@grpc/grpc-js": {
-      "version": "1.10.6",
-      "resolved": "https://registry.npmjs.org/@grpc/grpc-js/-/grpc-js-1.10.6.tgz",
-      "integrity": "sha512-xP58G7wDQ4TCmN/cMUHh00DS7SRDv/+lC+xFLrTkMIN8h55X5NhZMLYbvy7dSELP15qlI6hPhNCRWVMtZMwqLA==",
+      "version": "1.10.8",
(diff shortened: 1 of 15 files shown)


alexmuller (approved):
Nice, splitting this out into different bits in `lib/` makes it more clear.

**Question:** did you want to update the use of appInfo to the new changes you added in #1082?

**Non-blocking** but I am slightly confused by the move of `options.authorizationHeader` to `options.tracing.authorizationHeader` but then you've kept the same environment variable of `OPENTELEMETRY_AUTHORIZATION_HEADER` which suggests that we will be sharing it between traces and metrics?

rowanmanning:
> Question: did you want to update the use of appInfo to the new changes you added in https://github.com/Financial-Times/dotcom-reliability-kit/pull/1082?

Yep I'll do that in a later PR, didn't want to delay this one.

> Non-blocking but I am slightly confused by the move of options.authorizationHeader to options.tracing.authorizationHeader but then you've kept the same environment variable of OPENTELEMETRY_AUTHORIZATION_HEADER which suggests that we will be sharing it between traces and metrics?

I couldn't be bothered to try and change the name of that one yet. I may deprecate this environment variable and switch to `OPENTELEMETRY_TRACING_AUTHORIZATION_HEADER` in future though

PR #1077, opened 2024-06-10
feat: pull the release version from package.json

rowanmanning:
This makes our calculation of system version a lot more solid. This is going to be required for OpenTelemetry metrics, because any metric sent without a version will end up in the Dead Letter metrics. Because of this I want to give users of Reliability Kit more of a guarantee that their metrics will arrive.


rowanmanning:
Superseded by #1082

PR #1082, opened 2024-06-11
Broad app info improvements (required for OpenTelemetry metrics)

### LINKED ISSUE #818: Align app info properties with OpenTelemetry
OpenTelemetry have already done a lot of work around consolidating the different environment variables you find across different cloud platforms. We should lean on their work and group our app info properties in the same way.

We align more with a [well-defined standard](https://opentelemetry.io/docs/specs/semconv/) and tidy up the slight mess that is all of app info's top-level properties. We'll also have easy answers if we want to add future properties, we just go with whatever OpenTelemetry have done.

### Deprecate existing properties

Firstly we should leave the existing properties in place so as not to introduce a breaking change. Instead of removing them we should use getter methods and log deprecation warnings the first time one is used. [See the logger for an example](https://github.com/Financial-Times/dotcom-reliability-kit/blob/9359fdd042695626cf997aab95e00cf1d8170a8d/packages/logger/lib/logger.js#L320-L330) of how we've done this in the past.

We should deprecate the following properties:

  * `appInfo.cloudProvider`
  * `appInfo.commitHash`
  * `appInfo.processType`
  * `appInfo.region`
  * `appInfo.releaseDate`
  * `appInfo.releaseVersion`

I don't think we should deprecate these properties:

  * `appInfo.environment`: This is accessed really frequently and I think it deserves a shortcut
  * `appInfo.systemCode`: Similarly this ensures that system code is easy to find, even if it's an alias of a newer property

### Add new properties

We should add the following properties:

| New property                       | Old property     | Where from?                                               |
|------------------------------------|------------------|-----------------------------------------------------------|
| `service.name`                     | `systemCode`     | `SYSTEM_CODE` or `package.json` `name` property           |
| `service.version`                  | `releaseVersion` | `HEROKU_RELEASE_VERSION` or `AWS_LAMBDA_FUNCTION_VERSION` |
| `service.instance.id`              | -                | `HEROKU_DYNO_ID`                                          |
| `heroku.app.id`                    | -                | `HEROKU_APP_ID`                                           |
| `heroku.release.commit`            | `commitHash`     | `HEROKU_SLUG_COMMIT`                                      |
| `heroku.release.creationTimestamp` | `releaseDate`    | `HEROKU_RELEASE_CREATED_AT`                               |
| `cloud.provider`                   | `cloudProvider`  | conditionally set based on presence of env vars           |
| `cloud.region`                     | `region`         | `AWS_REGION`, `AWS_DEFAULT_REGION`, or `REGION`           |
| `deployment.environment`           | `environment`    | `NODE_ENV`                                                |

## Alternatives

We could do nothing, but we'll keep cluttering up the top level properties the more we add. We'll also have to have more conversations about naming things.

rowanmanning:
These are all required to add OpenTelemetry metrics, we need to solidify the app info properties that we rely on otherwise our metrics are rejected. This does the following:

  * Firms up the `releaseVersion` property, required in our metrics
  * Adds in an `instanceId` property, required in our metrics
  * Adds aliases for OpenTelemetry semantic conventions, which is the first step towards adopting them fully. We'll deprecate old property names in future (resolves #818)

### DIFF
--- a/packages/app-info/README.md
+++ b/packages/app-info/README.md
@@ -116,6 +116,8 @@ This object contains aliases for the main `appInfo` properties that correspond t
   * `appInfo.semanticConventions.service.version` aliases `appInfo.releaseVersion`
   * `appInfo.semanticConventions.service.instance.id` aliases `appInfo.instanceId`
 
+> [!WARNING]
+> While all other properties default to `null` if they can't be calculated, the semantic conventions properties default to `undefined`. This is to ensure better compatibility with OpenTelemetry SDKs.
 
 ## Migrating
 
--- a/packages/app-info/README.md
+++ b/packages/app-info/README.md
@@ -12,8 +12,9 @@ A utility to get application information (e.g. the system code) in a consistent 
     * [`appInfo.systemCode`](#appinfosystemcode)
     * [`appInfo.processType`](#appinfoprocesstype)
     * [`appInfo.cloudProvider`](#appinfocloudprovider)
-    * [`appInfo.herokuAppId`](#appinfoherokuappId)
-    * [`appInfo.herokuDynoId`](#appinfoherokudynoId)
+    * [`appInfo.herokuAppId`](#appinfoherokuappid)
+    * [`appInfo.herokuDynoId`](#appinfoherokudynoid)
+    * [`appInfo.semanticConventions`](#appinfosemanticconventions)
   * [Migrating](#migrating)
   * [Contributing](#contributing)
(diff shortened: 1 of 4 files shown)


alexmuller (approved):
Nice, looks good.

When `appInfo.instanceId` is a random UUID it is reused across the instantiation of `appInfo` which is the behaviour we want - in my opinion. Every application restart will create a new UUID but the UUID will be consistent between restarts.

PR #1089, opened 2024-06-13
Add OpenTelemetry metrics

rowanmanning:
This adds metrics to the OpenTelemetry package, allowing apps using Reliability Kit to send metrics via the FT's metrics collector. Under the hood this essentially runs the same code as is outlined [in the Tech Hub guide](https://tech.in.ft.com/tech-topics/observability/opentelemetry/heroku) but will less boilerplate required for an app.

## Basic setup

Instead of copy/pasting example code, you should now be able to do the following:

Install the Reliability Kit `opentelemetry` package:

```
npm install @dotcom-reliability-kit/opentelemetry
```

Add it to your startup script:

```
node --require @dotcom-reliability-kit/opentelemetry/setup ./my-app.js
```

Add some environment variables to your app (see the [internal collector repo](https://github.com/Financial-Times/observability/tree/main/services/otel-gateway-collector#readme) for these values, this is a public repo so I don't want to share them in this PR description):

```
OPENTELEMETRY_METRICS_ENDPOINT=<value>
OPENTELEMETRY_API_GATEWAY_KEY=<value>
```

> [!TIP]
> Note that the environment variable names are different to the ones in the Tech Hub guides!

(see the README for more ways to set up)

## Custom metrics

As outlined in the updated README, we alias some OpenTelemetry methods to make it easier to send custom metrics (easier as in you don't also need to install the related `@opentelemetry/...` packages separately):

```js
const { getMeter } = require('@dotcom-reliability-kit/opentelemetry');

const meter = getMeter('my-app');
const hitCounter = meter.createCounter('my-app.hits');

// Assumes that `app` is an Express application instance
app.get('/', (request, response) => {
    hitCounter.add(1);
    response.send('Thanks for visiting');
});
```

## Testing this code locally

I've been testing locally using [Reliability Dashboards](https://github.com/Financial-Times/reliability-dashboards) which already has the `@dotcom-reliability-kit/opentelemetry` package installed. Clone the repo, install dependencies, and then link to a local copy of the package on this branch:

```
rm -rf node_modules/@dotcom-reliability-kit/opentelemetry
ln -s <RELIABILITY_KIT_DIR>/packages/opentelemetry node_modules/@dotcom-reliability-kit/opentelemetry
```

You can then start the app, passing in environment variables:

```
OPENTELEMETRY_API_GATEWAY_KEY=XXXXXX OPENTELEMETRY_METRICS_ENDPOINT=XXXXXX npm start
```

You should soon see metrics in Grafana in the `OpenTelemetry Test` data source.

---

See-also: [CPREL-1087](https://financialtimes.atlassian.net/browse/CPREL-1087)

[CPREL-1087]: https://financialtimes.atlassian.net/browse/CPREL-1087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -152,6 +152,8 @@ opentelemetry.setup({ /* ... */ });
     </tr>
 </table>
 
+This method returns any SDK instances created during setup. Calling this method a second time will return the same instances without rerunning setup.
+
 ### Running in production
 
 #### Production metrics
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -276,7 +276,7 @@ A URL to send OpenTelemetry metrics to. E.g. `http://localhost:4318/v1/metrics`.
 
 #### `options.metrics.apiGatewayKey`
 
-Set the `Authorization` HTTP header in requests to the central API-Gateway-backed OpenTelemetry metrics collector. Defaults to `undefined`.
+Set the `X-OTel-Key` HTTP header in requests to the central API-Gateway-backed OpenTelemetry metrics collector. Defaults to `undefined`.
 
 **Environment variable:** `OPENTELEMETRY_API_GATEWAY_KEY`<br/>
 **Option:** `metrics.apiGatewayKey` (`String`)
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -30,6 +30,7 @@ An [OpenTelemetry](https://opentelemetry.io/docs/what-is-opentelemetry/) client 
     * [`options.tracing.authorizationHeader`](#optionstracingauthorizationheader)
     * [`options.tracing.samplePercentage`](#optionstracingsamplepercentage)
     * [`OTEL_` environment variables](#otel_-environment-variables)
+* [Migrating](#migrating)
 * [Contributing](#contributing)
 * [License](#license)
 
@@ -117,9 +118,9 @@ OpenTelemetry will be [configured](#configuration-options) with environment vari
 If you'd like to customise the OpenTelemetry config more and have control over what runs, you can include in your code:
 
 ```js
-import setupOpenTelemetry from '@dotcom-reliability-kit/opentelemetry';
+import * as opentelemetry from '@dotcom-reliability-kit/opentelemetry';
 // or
-const setupOpenTelemetry = require('@dotcom-reliability-kit/opentelemetry');
+const opentelemetry = require('@dotcom-reliability-kit/opentelemetry');
 ```
 
 Call the function, passing in [configuration options](#configuration-options):
@@ -128,7 +129,7 @@ Call the function, passing in [configuration options](#configuration-options):
 > This **must** be the first function called in your application for OpenTelemetry to be set up correctly (including before other `import`/`require` statements).
 
 ```js
-setupOpenTelemetry({ /* ... */ });
+opentelemetry.setup({ /* ... */ });
 ```
 
 <table>
@@ -225,7 +226,7 @@ EXAMPLE=true npm start
 For the [manual setup](#manual-setup), you'll need to use an options object, e.g.
 
 ```js
-setupOpenTelemetry({
+opentelemetry.setup({
     example: true
 });
 ```
@@ -283,6 +284,11 @@ OpenTelemetry itself can be configured through `OTEL_`-prefixed environment vari
 
 > [!CAUTION]
 > We strongly advise against using these. The power of this module is consistency and any application-specific changes should be considered. If you use these environment variables we won't offer support if things break.
+
+
+## Migrating
+
+Consult the [Migration Guide](./docs/migration.md) if you're trying to migrate to a later major version of this package.
 
 
 ## Contributing
--- a/packages/opentelemetry/README.md
+++ b/packages/opentelemetry/README.md
@@ -11,6 +11,7 @@ An [OpenTelemetry](https://opentelemetry.io/docs/what-is-opentelemetry/) client 
     * [Automated setup with `--require`](#automated-setup-with---require)
     * [Automated setup with `require()`](#automated-setup-with-require)
     * [Manual setup](#manual-setup)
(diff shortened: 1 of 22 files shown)


packages/opentelemetry/README.md
```diff
@@ -212,6 +263,24 @@ setupOpenTelemetry({
...
+A URL to send OpenTelemetry metrics to. E.g. `http://localhost:4318/v1/metrics`. Defaults to `undefined` which means that OpenTelemetry metrics will not be sent.
+
+**Environment variable:** `OPENTELEMETRY_METRICS_ENDPOINT`<br/>
+**Option:** `metrics.endpoint` (`String`)
+
+#### `options.metrics.apiGatewayKey`
+
+Set the `Authorization` HTTP header in requests to the central API-Gateway-backed OpenTelemetry metrics collector. Defaults to `undefined`.
+
+**Environment variable:** `OPENTELEMETRY_API_GATEWAY_KEY`<br/>
```
rowanmanning, line 281: **question:** I wondered if it was worth naming this `OPENTELEMETRY_METRICS_API_GATEWAY_KEY`. It really depends whether we're likely to be able to use the same key in future if tracing support is added centrally. I imagine the answer is "not sure" but @sjparkinson do you think metrics/tracing will require different keys in future?

packages/opentelemetry/lib/config/metrics.js
```diff
@@ -0,0 +1,62 @@
...
+			endpoint: options.endpoint
+		});
+	} else {
+		logger.info({
+			event: 'OTEL_METRICS_STATUS',
+			message:
+				'OpenTelemetry metrics are disabled because no metrics endpoint was set',
+			enabled: false,
+			endpoint: null
+		});
```
rowanmanning, line 58: **note:** I decided to switch this out from a warning because it could be a bit spammy. I think if metrics aren't being received then it's up to our Grafana config to alert rather than having something in the logs. Something I could add is an error log if:

* The OpenTelemetry package is installed
* Both tracing _and_ metrics endpoints are not configured
* The environment is `production`

This feels like a definite misconfiguration and might help alert us to badly configured apps?

packages/opentelemetry/lib/index.js
```diff
@@ -13,20 +18,44 @@ const logger = require('@dotcom-reliability-kit/logger');
...
+ *      Returns any created SDK instances.
  */
 function setupOpenTelemetry({
 	authorizationHeader,
+	metrics: metricsOptions,
 	tracing: tracingOptions
 } = {}) {
+	if (instances) {
+		return instances;
+	}
```
rowanmanning, line 57: **note:** I know that OpenTelemetry itself has some checks for this kind of thing, but I wanted to be sure that accidentally setting up otel twice doesn't cause weird issues. This also allows us to check whether OpenTelemetry has been set up via Reliability Kit before allowing other operations.

packages/opentelemetry/README.md
sjparkinson: The same key and endpoint will be used for traces when we eventually get around to supporting that centrally.

sjparkinson: `OTEL_EXPORTER_OTLP_API_KEY` is what we have used in docs to align with the naming convention of other environment variables used by the SDKs.

Strictly speaking we could/should? be using [`OTEL_EXPORTER_OTLP_HEADERS`](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_headers), but the format of the value doesn't work well with the API Gateway's Doppler integration.

rowanmanning: Ah yeah about those, we're warning people off them because they can mess with the way we've configured everything. It's fine to use `OTEL_` vars but we don't want to have to support you if they break something 😂

https://github.com/Financial-Times/dotcom-reliability-kit/blob/d49b41d02292d7d0a1201d43fa406f0355500936/packages/opentelemetry/lib/index.js#L59-L69

This is why we went with our own naming instead of the one in your docs.

packages/opentelemetry/lib/config/resource.js
```diff
@@ -1,25 +1,27 @@
...
 	// We set OpenTelemetry resource attributes based on app data
 	return new Resource({
-		[SEMRESATTRS_SERVICE_NAME]: appInfo.systemCode || undefined,
-		[SEMRESATTRS_SERVICE_VERSION]: appInfo.releaseVersion || undefined,
-		[SEMRESATTRS_CLOUD_PROVIDER]: appInfo.cloudProvider || undefined,
-		[SEMRESATTRS_CLOUD_REGION]: appInfo.region || undefined,
-		[SEMRESATTRS_DEPLOYMENT_ENVIRONMENT]: appInfo.environment || undefined
+		[SEMRESATTRS_SERVICE_NAME]: appInfo.service.name,
+		[SEMRESATTRS_SERVICE_VERSION]: appInfo.service.version,
+		[SEMRESATTRS_SERVICE_INSTANCE_ID]: appInfo.service.instance.id,
```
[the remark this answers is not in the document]
rowanmanning: Yep we set a different ID per-process 👍 this value comes from the app-info package and we default to a UUID:

https://github.com/Financial-Times/dotcom-reliability-kit/blob/ad0245d25db9dfe60ba0cc0834e461183c5b7320/packages/app-info/lib/index.js#L153-L159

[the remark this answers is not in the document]
rowanmanning: Ah right wait I see, so you think we should drop the `HEROKU_DYNO_ID` too?

packages/opentelemetry/lib/config/metrics.js
rowanmanning: Oh that's interesting though, it'd definitely be possible for this library to just send a new metric on a timer while the process is running, might be a bit spammy though? If we did something like that we'd probably want to discuss naming etc

alexmuller (approved):
👍

Issue #946, opened 2024-02-21
Write manual type declarations for all modules


rowanmanning:
In #927 we added the ability for a module to have manually-written type declarations that are committed to the repo. We need to migrate all the remaining modules. This mega-issue will keep track of progress and contain notes on the process.

  - [x] app-info
  - [x] crash-handler
  - [x] errors
  - [x] eslint-config
  - [x] fetch-error-handler
  - [x] log-error
  - [x] middleware-log-errors
  - [x] middleware-render-error-info
  - [x] opentelemetry
  - [x] serialize-error
  - [x] serialize-request
  - [x] logger

PR #1106, opened 2024-06-28
Even more manual types

rowanmanning:
We're so close to completing #946, this adds types for all the remaining packages except logger which is the most complex.


CyntiBinti (approved):
🙌🏽  Great job powering through this! One step closer

PR #1107, opened 2024-07-01
Add the final manual type declarations

### LINKED ISSUE #946: Write manual type declarations for all modules

rowanmanning:
This adds type declarations to the final package: it's the most complex but I decided to just replicate the existing generated types. Because we no longer need a build step I'm removing all of the build-related tooling and updating the documentation accordingly.

This unlocks the ability to migrate to TypeScript 5.5 which is also included in this PR. It may be easier to review each commit separately because the second one moves types into `@import` statements across the board.

This resolves #946.


CyntiBinti (approved):
Woohoo 🙌🏽 🚀

PR #1112, opened 2024-07-02
feat: allow setting the deployment environment

### LINKED ISSUE #1111: We don't differentiate between deployment environment and Node.js environment
In [app-info](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/app-info#readme) (and, by extension, [opentelemetry](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/opentelemetry#readme)) we don't have a way of differentiating between the deployment environment (e.g. `development`, `staging`, `production`) and the Node.js environment (the value of `NODE_ENV`). This means that we can't tell the difference between production and staging metrics, because in both cases the `NODE_ENV` environment variable is set to `production`.

**Node.js version:** All
**Impacted package(s):** [app-info](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/app-info#readme), [opentelemetry](https://github.com/Financial-Times/dotcom-reliability-kit/tree/main/packages/opentelemetry#readme)

## Steps to reproduce

This isn't very easy to reproduce as you need to set up a system in OpenTelemetry that has a long-lived staging environment. It's easier to highlight the code in question.

  * [Here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/016d672f27e8458bf65a6a94a9338c24cbc96180/packages/app-info/lib/index.js#L88) you can see that we use `process.env.NODE_ENV` for `appInfo.environment`

  * [Here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/016d672f27e8458bf65a6a94a9338c24cbc96180/packages/app-info/lib/index.js#L173-L175) we use that same variable for `semanticConventions.deployment.environment`

  * [Here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/016d672f27e8458bf65a6a94a9338c24cbc96180/packages/opentelemetry/lib/config/resource.js#L25) we use this alias to set the OpenTelemetry `deployment.environment` resource attribute.

  * Check staging vs production environment variables for one of your apps. The `NODE_ENV` environment variable is almost certainly set to `production` in both

  * It's not possible to set the `NODE_ENV` environment variable to something like `staging` because a lot of our underlying tooling changes behaviour based on the value being `production` (e.g. [Express caching rendered views](https://github.com/expressjs/express/blob/4cf7eed927d3ccd3f1d0c9a14d562ec0a1635e86/lib/application.js#L125-L127)).

## Expected behaviour

I expect us to be able to set the [deployment environment](https://opentelemetry.io/docs/specs/semconv/attributes-registry/deployment/) to something other than `NODE_ENV` which has a different purpose.

We do need this value in `app-info` to ensure that our logs and metrics have the same value, and to make this available to apps that want to detect whether they're running in production or staging. Because of this, I don't think using the [`OTEL_RESOURCE_ATTRIBUTES` environment variable](https://opentelemetry.io/docs/concepts/resources/#custom-resources) is enough.

[Unfortunately there are no built-in resource detectors](https://github.com/search?q=org%3Aopen-telemetry+DEPLOYMENT_ENVIRONMENT+language%3ATypeScript&type=code) for this value _except_ for Microsoft Azure which uses the `WEBSITE_SLOT_NAME` environment variable. I don't think this makes sense for us.

Across the FT we use a few different environment variables:

  * [`RELEASE_ENV`](https://github.com/search?q=org%3AFinancial-Times%20release_env&type=code) is fairly common
  * [`DEPLOY_STAGE`](https://github.com/search?q=org%3AFinancial-Times%20deploy_stage&type=code) is uncommon but is closer to the semantic resource attribute name
  * ... there may be more, please let me know!

As we haven't settled on a value, I'd be tempted to try and standardise on a more descriptive one, e.g. `DEPLOYMENT_ENVIRONMENT` or `DEPLOYMENT_ENV`.

It'd be possible to cater for all of these using the following:

```js
const environment =
	process.env.DEPLOYMENT_ENVIRONMENT ||
	process.env.DEPLOY_STAGE ||
	process.env.RELEASE_ENV ||
	process.env.NODE_ENV ||
	'development';
```

I think this is what I'm leaning towards as a solution.

rowanmanning:
We're now looking at more different environment variables to determine the deployment environment the app is running in. `NODE_ENV` by itself isn't sufficient because it's nearly always set to `production` when running on a server, even for staging apps.

This resolves #1111.

### DIFF
--- a/packages/app-info/README.md
+++ b/packages/app-info/README.md
@@ -49,7 +49,12 @@ For AWS Lambda, you can use a plugin like [serverless-plugin-git-variables](http
 
 ### `appInfo.environment`
 
-Get the application environment, normally either `development` or `production`. This will be a string, using `process.env.NODE_ENV` and defaulting to `development`.
+Get the application environment, normally either `development` or `production`. This will be a string, trying each of the following environment variables in order and defaulting to `development`:
+
+  * `DEPLOYMENT_ENVIRONMENT`
+  * `RELEASE_ENV`
+  * `ENVIRONMENT`
+  * `NODE_ENV`
 
 ### `appInfo.region`
 
--- a/packages/app-info/lib/index.js
+++ b/packages/app-info/lib/index.js
@@ -80,12 +80,17 @@ exports.commitHash =
 	null;
 
 /**
- * The application environment.
+ * The application deployment environment.
  *
  * @readonly
  * @type {string}
  */
-exports.environment = process.env.NODE_ENV || 'development';
+exports.environment =
+	process.env.DEPLOYMENT_ENVIRONMENT ||
+	process.env.RELEASE_ENV ||
+	process.env.ENVIRONMENT ||
+	process.env.NODE_ENV ||
+	'development';
 
 /**
  * The region the application is running in.
--- a/packages/app-info/test/unit/lib/index.spec.js
+++ b/packages/app-info/test/unit/lib/index.spec.js
@@ -13,7 +13,10 @@ describe('@dotcom-reliability-kit/app-info', () => {
 		process.env.HEROKU_RELEASE_CREATED_AT = 'mock-heroku-release-date';
 		process.env.HEROKU_RELEASE_VERSION = 'mock-heroku-release-version';
 		process.env.HEROKU_SLUG_COMMIT = 'mock-heroku-commit-hash';
-		process.env.NODE_ENV = 'mock-environment';
+		process.env.DEPLOYMENT_ENVIRONMENT = 'mock-deployment-environment';
+		process.env.RELEASE_ENV = 'mock-release-env';
+		process.env.ENVIRONMENT = 'mock-environment';
+		process.env.NODE_ENV = 'mock-node-env';
 		process.env.REGION = 'mock-region';
 		process.env.SYSTEM_CODE = 'mock-system-code';
 		process.env.DYNO = 'mock-heroku-process-type.1';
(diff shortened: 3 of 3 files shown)


packages/app-info/README.md
```diff
@@ -49,7 +49,12 @@ For AWS Lambda, you can use a plugin like [serverless-plugin-git-variables](http
 
 ### `appInfo.environment`
 
-Get the application environment, normally either `development` or `production`. This will be a string, using `process.env.NODE_ENV` and defaulting to `development`.
+Get the application environment, normally either `development` or `production`. This will be a string, trying each of the following environment variables in order and defaulting to `development`:
```
rowanmanning, line 52: Right now we're not doing any normalisation here but I'm not against adding this in future if you want consistent values, @sjparkinson? E.g. we could convert `p` and `prod` to `production` automatically

CyntiBinti (approved):

packages/app-info/README.md
sjparkinson: Not fussed! Great if it's aligned but we could also do that in the Gateway Collector if we want consistency across the board.

sjparkinson: As a note, the values the Gateway supports for "production are as you listed, `p`, `prod` or `production` nothing else.

https://github.com/Financial-Times/observability/blob/aeb21e682ebfcaefbefa6501705c9645a523c130/services/otel-gateway-collector/config/otel-collector-pipeline-otlp.yaml#L42-L47

PR #1123, opened 2024-07-10
chore: simplify the dependabot config

rowanmanning:
Dependabot introduced a `directories` key a while ago which means you can use a single config for all packages in a monorepo instead of having to repeat the config many times.


alexmuller (approved):
Looks good, docs for creating a new package are still valid.

PR #1125, opened 2024-07-10
Reduce OpenTelemetry warnings

rowanmanning:
This reduces warnings from the OpenTelemetry package:

- **Disable the pino instrumentation:**

    The pino instrumentation is causing us issues because we import our logger before the instrumentation can happen. It's logging an obnoxious warning. I think this is fine to disable because:
    
    1. We use pino to log to stdout so we're not missing a lot of useful info about the way it runs
    
    2. The instrumentation only adds traces, not metrics. We run pino async so it doesn't block further execution and shouldn't impact the timings of the routes etc that use it
    
    3. By default, this instrumentation tries to use the OpenTelemetry logs SDK. I'd rather not load some stuff that we don't want to use.

- **Wrap the Reliability Kit logger:**

    We have been losing information from the OpenTelemetry internal logs because they log multiple messages in a row which pino/reliability kit does not support. We don't support this because we want to maintain consistency with n-logger.
    
    To fix this, we need to wrap the Reliability Kit logger and extract any extra logged information into a new property. I've gone with `details`. Testing this locally, it's not common to get any info but it's important when we do get it - it helped me debug the issue with loading Pino ahead of instrumentation.

- **Manually provide a meter to `HostMetrics`**

    This suppresses an annoying warning. The host metrics class already gets the global meter provider in the same way but logs a warning if it has to. This is a lot of noise in our logs.

What it doesn't do:

- We're not handling `Accessing resource attributes before async attributes settled`, we'll wait to see [what happens with this issue](https://github.com/open-telemetry/opentelemetry-js/issues/4638)

### DIFF
--- a/packages/opentelemetry/lib/config/instrumentations.js
+++ b/packages/opentelemetry/lib/config/instrumentations.js
@@ -28,6 +28,9 @@ exports.createInstrumentationConfig = function createInstrumentationConfig() {
 			},
 			'@opentelemetry/instrumentation-fs': {
 				enabled: false
+			},
+			'@opentelemetry/instrumentation-pino': {
+				enabled: false
 			}
 		}),
 		new RuntimeNodeInstrumentation()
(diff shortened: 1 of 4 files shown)


CyntiBinti (approved):
Thanks for clarifying the otels internal logging process 🙇🏽

PR #1126, opened 2024-07-11
feat: add a `logUserErrorsAsWarnings` option

rowanmanning:
This adds a `logUserErrorsAsWarnings` option which allows you to change the log level for handled errors. If this option is set to `true` then:

  * Errors with a `status` or `statusCode` property of `400`-`499` will be logged with a level of "warn"

  * All other errors will still be logged with a level of "error"

I think in future we'll switch this to be the default behaviour and remove the option.


CyntiBinti (approved):
A lovely feature!

Issue #945, opened 2024-02-21
Remove our use of default exports.


rowanmanning:
Default exports cause a lot of issues with the package types, we should stop using them and use named exports across any Reliability Kit modules we expect to be imported (i.e. all except `eslint-config`). This would be a breaking change.

## What problem does this feature solve?

Default exports cause many of the compatibility issues between CommonJS and ESM. They also require us to make adjustments to get the types to work properly. If we weren't using default exports then we'd sidestep some of the issues with TypeScript and [`esModuleInterop`](https://www.typescriptlang.org/tsconfig#esModuleInterop), effectively reducing the number of module systems we test against by two.

If we stop using default exports then it's also a lot easier to start writing our code as ESM and cross-compiling to CommonJS (something I've experimented with but got blocked by our default exports). Writing as ESM feels more future-proof.

## Ideal solution

We move all modules to used named exports, which would require code changes in downstream apps like this:

```diff
- const registerCrashHandler = require('@dotcom-reliability-kit/crash-handler');
+ const { registerCrashHandler } = require('@dotcom-reliability-kit/crash-handler');

- import registerCrashHandler from '@dotcom-reliability-kit/crash-handler';
+ import { registerCrashHandler } from '@dotcom-reliability-kit/crash-handler';
```

Although it requires a code change, it's very find/replaceable and we could do it with codemods (something that the Developer Automation team have been looking into).

If we think it's worthwhile, we could also introduce named exports to the modules as aliases in the short-term so that teams can start migrating ahead of the breaking change.

## Alternatives

We could continue to use default exports but we'll be making it harder to migrate to ESM and we'd be accepting continued maintenance of several module compatibility tests that we could otherwise remove.

We could also start doing this work earlier for our low-level modules, e.g. `app-info`, `serialize-request`, and `serialize-error` – these modules are used by apps less frequently and mostly used in our own higher-level modules.

rowanmanning:
Yep I don't think we need to do this any more with the migration to ESM. CommonJS users will have to reference `default` which is enough of a breaking change anyway and users of ESM will be able to work with our modules normally. Great!

PR #1325, opened 2025-02-04
feat: manage bad requests with 'allow request methods' package

CyntiBinti:
Current behaviour: a DDoS can use an unexpected HTTP method to bypass the Fastly cache and cause interruption to service. For example, making a request to the homepage as a POST rather than a GET. Express will return a 404 for this error.

Ideal future work: If we can transform this 404 error into a 405 Method Not Allowed, we may be able to use Signal Sciences to block this kind of attack very quickly. This is because it’s highly unlikely that a genuine user will accidentally make a POST request.

Definition of done for this ticket: Build and release a package containing an Express middleware that applications can consume which will cause unexpected request methods to error with a 405 rather than a 404.

Inspo: https://github.com/rowanmanning/allow-methods/blob/main/index.js

To-Do:

- [x] Write tests

See Also: [CPREL-1276]

[CPREL-1276]: https://financialtimes.atlassian.net/browse/CPREL-1276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/middleware-allow-request-methods/README.md
+++ b/packages/middleware-allow-request-methods/README.md
@@ -6,12 +6,9 @@ Express middleware that returns 405 (rather than 404) for disallowed request met
   * [Usage](#usage)
     * [Configuration options](#configuration-options)
       * [`options.allowedMethods`](#optionsallowedmethods)
-      * [`options.message`](#optionsmessage)
-      * [`options.logger`](#optionslogger)
   * [Migrating](#migrating)
   * [Contributing](#contributing)
   * [License](#license)
-
 
 ## Usage
 
@@ -24,30 +21,45 @@ npm install --save @dotcom-reliability-kit/middleware-allow-request-methods
 Include in your code:
 
 ```js
-import allowRequestMethods from '@dotcom-reliability-kit/middleware-allow-request-methods';
+import { allowRequestMethods } from '@dotcom-reliability-kit/middleware-allow-request-methods';
 // or
-const allowRequestMethods = require('@dotcom-reliability-kit/middleware-allow-request-methods');
+const { allowRequestMethods } = require('@dotcom-reliability-kit/middleware-allow-request-methods');
 ```
+
+We recommend always using this middleware globally with app.use as a first middleware in your app. This is because, if a bad actor is making requests to your app to find attack vectors, you throw their request out as early as possible.
+
+Route-specific blocking of methods is an additional layer of protection you can explore. It may be that your app does support POST requests for a form but the main view is GET only. You can filter out further junk requests on a per-route basis by using the app.route('...').all() method or use with a path.
 
 Example usage:
 
 ```js
 const express = require('express');
-const allowRequestMethods = require('@dotcom-reliability-kit/middleware-allow-request-methods');
+const { allowRequestMethods } = require('@dotcom-reliability-kit/middleware-allow-request-methods');
 
 const app = express();
 
-// Apply the middleware to specific routes, for example:
-app.use('/', allowRequestMethods(['GET'])); // Allow only GET requests on '/'
-app.use('/submit', allowRequestMethods(['POST'])); // Allow only POST requests on '/submit'
+// Allow only certain request methods for the entire app. If you're
+// doing this, it must be above ALL routes you want it to apply to:
+app.use(allowRequestMethods({ allowedMethods: ['GET', 'HEAD', 'POST'] }));
 
-// Define your routes
-app.get('/', (req, res) => {
-  res.send('Homepage');
-});
+// Allow only certain request methods for a specific route, e.g. here
+// we only allow `GET` and `HEAD` methods for the home page. Note that
+// we have to use `all` for the allowed methods here THEN define the get
+// request handler:
+app
+ .route('/')
+ .all(allowRequestMethods({ allowedMethods: ['GET', 'HEAD'] }))
+ .get((request, response) => {
+  response.send('Homepage');
+ });
 
-app.post('/submit', (req, res) => {
-  res.send('Form submitted');
+// You can also allow methods for a subset of routes. Remember that this
+// applies for all routes that START with the value. E.g. the following
+// will also only allow POST requests on `/submit/example`:
+app.use('/submit', allowRequestMethods({ allowedMethods: ['POST'] }));
+
+app.post('/submit', (request, response) => {
+ response.send('Form submitted');
 });
 
 app.listen(3000, () => console.log('Server running on port 3000'));
@@ -69,31 +81,9 @@ An array of HTTP methods that are allowed for the route. This must be an `Array`
 
 This option defaults to `[]`.
 
-#### `options.message`
-
-A string to be used as the response body when a request is made with an unsupported method.
-
-This option defaults to `'Method Not Allowed'`.
-
-#### `options.logger`
-
-A logger object which implements two methods, `error` and `warn`, which have the following permissive signature:
-
-```ts
-type LogMethod = (...logData: any) => any;
-```
-
-This is passed directly onto the relevant log-error method, [see the documentation for that package for more details](../log-error/README.md#optionslogger).
-
-## Migrating
-
-Consult the [Migration Guide](./docs/migration.md) if you're trying to migrate to a later major version of this package.
-
-
 ## Contributing
 
 See the [central contributing guide for Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md).
-
 
 ## License
 
--- a/packages/middleware-allow-request-methods/README.md
+++ b/packages/middleware-allow-request-methods/README.md
@@ -1,0 +1,101 @@
+
+# @dotcom-reliability-kit/middleware-allow-request-methods
+
+Express middleware that returns 405 (rather than 404) for disallowed request methods. This module is part of [FT.com Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit#readme).
+
+  * [Usage](#usage)
+    * [Configuration options](#configuration-options)
+      * [`options.allowedMethods`](#optionsallowedmethods)
+      * [`options.message`](#optionsmessage)
+      * [`options.logger`](#optionslogger)
+  * [Migrating](#migrating)
+  * [Contributing](#contributing)
+  * [License](#license)
+
+
+## Usage
+
+Install `@dotcom-reliability-kit/middleware-allow-request-methods` as a dependency:
+
+```bash
+npm install --save @dotcom-reliability-kit/middleware-allow-request-methods
+```
+
+Include in your code:
+
+```js
+import allowRequestMethods from '@dotcom-reliability-kit/middleware-allow-request-methods';
+// or
+const allowRequestMethods = require('@dotcom-reliability-kit/middleware-allow-request-methods');
+```
+
+Example usage:
+
+```js
+const express = require('express');
+const allowRequestMethods = require('@dotcom-reliability-kit/middleware-allow-request-methods');
+
+const app = express();
+
+// Apply the middleware to specific routes, for example:
+app.use('/', allowRequestMethods(['GET'])); // Allow only GET requests on '/'
+app.use('/submit', allowRequestMethods(['POST'])); // Allow only POST requests on '/submit'
+
+// Define your routes
+app.get('/', (req, res) => {
+  res.send('Homepage');
+});
+
+app.post('/submit', (req, res) => {
+  res.send('Form submitted');
+});
+
+app.listen(3000, () => console.log('Server running on port 3000'));
+```
+
+### Configuration options
+
+Config options can be passed into the `allowRequestMethods` function as an object with any of the keys below.
+
+```js
+app.use(allowRequestMethods({
+    // Config options go here
+}));
+```
+
+#### `options.allowedMethods`
+
+An array of HTTP methods that are allowed for the route. This must be an `Array` of `String`s, with each string being an HTTP method. It's important that you do not include methods which are not supported by the route.
+
+This option defaults to `[]`.
+
+#### `options.message`
+
+A string to be used as the response body when a request is made with an unsupported method.
+
+This option defaults to `'Method Not Allowed'`.
+
+#### `options.logger`
+
+A logger object which implements two methods, `error` and `warn`, which have the following permissive signature:
+
+```ts
+type LogMethod = (...logData: any) => any;
+```
+
+This is passed directly onto the relevant log-error method, [see the documentation for that package for more details](../log-error/README.md#optionslogger).
+
+## Migrating
+
+Consult the [Migration Guide](./docs/migration.md) if you're trying to migrate to a later major version of this package.
+
+
+## Contributing
+
+See the [central contributing guide for Reliability Kit](https://github.com/Financial-Times/dotcom-reliability-kit/blob/main/docs/contributing.md).
+
+
+## License
+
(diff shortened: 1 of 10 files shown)


packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+			// If the allowed methods array is empty, you can either allow all methods or reject everything
+			if (normalisedAllowedRequestMethods.length === 0) {
+				// TODO: Option 1: Allow all methods (no restriction) i.e. request proceeds as normal
+				return next();
+
+				// TODO: or Option 2: Reject all methods (405 for every request) i.e. block all requests when no methods are explicitly stated
+				// response.header('Allow', normalisedAllowedRequestMethods.join(', '));
+				// response.status(405).send(options.message);
+				// return next(new MethodNotAllowedError(options.message));
+			}
```
CyntiBinti, line 48: Had a thought on the intention of this function and felt it could either be option 1 or option 2, but want to double check, as I initially wrote it assuming option 2 (i.e. I'm assuming if someone's using this package then they're using it with the intention of it's stated purpose and thus they'll explicitly state the allowed methods for that route, else why would they be requiring this package into their app 🤷🏽‍♀️ )

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+	const normalisedAllowedRequestMethods = normaliseAllowedRequestMethods(
+		options.allowedMethods || []
+	);
+
+	return function allowRequestMethodsMiddleware(
+		error,
+		request,
+		response,
+		next
+	) {
```
rowanmanning, line 32: **issue:** we don't want this to be an Express error handler because otherwise it won't ever fire unless an error's already been thrown by a previous middleware. We want a regular handler which only accepts three arguments: `request`, `response`, and `next`. This is different to the other middleware we've written which both _do_ have to handle errors in the request/response cycle

packages/middleware-allow-request-methods/lib/index.js
rowanmanning: Yep I agree with your gut here - option 1 should not be a use-case, if you don't want to protect an endpoint then don't use this middleware. I think we should go one further and throw a `TypeError` if no allowed methods are specified. We do similar option validation in other places, [e.g. here](https://github.com/Financial-Times/dotcom-reliability-kit/blob/d2058bfdc140b8144a080a32865b57de08bc8869/packages/middleware-log-errors/lib/index.js#L34-L38)

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+	 * @type {string}
+	 */
+	name = 'MethodNotAllowedError';
+
+	/** @type {number} */
+	status = 405;
+
+	/** @type {number} */
+	statusCode = 405;
+}
```
rowanmanning, line 109: **suggestion:** there's no need for this - we should be using Reliability Kit's errors because we get a lot of extra information for free and we don't need to maintain a new class. My suggestion would be:

```js
// Imports...
const { UserInputError } = require('@dotcom-reliability-kit/errors');

// In your middleware...
next(new UserInputError({ statusCode: 405 }));
```

There's a bunch of stuff we get for free here. The default message will be set based on the status code (so the standard `Method Not Allowed`. The default error `code` will be set to `HTTP_405` which allows teams to do more with the error later and we'll be able to see it in Splunk more easily.

You could also pass in a custom message however I don't think we should allow that as a v1 - it's extra code that we don't need yet and we can easily make it customisable later if somebody asks for it.

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+module.exports = allowRequestMethods;
+module.exports.default = module.exports;
```
rowanmanning, line 112: **issue:** we're trying to phase out the use of default exports (see #945) so for new packages we should always use named exports, otherwise we'll have to release a breaking change later. The docs will need updating too, but something like this matches [fetch-error-handler](https://github.com/Financial-Times/dotcom-reliability-kit/blob/d2058bfdc140b8144a080a32865b57de08bc8869/packages/fetch-error-handler/lib/index.js#L3-L4) (our current newest package):

```suggestion
exports.allowRequestMethods = allowRequestMethods;
```

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+				// return next(new MethodNotAllowedError(options.message));
+			}
+
+			// If the incoming request method is not in the allowed methods array, then send a 405 error
+			if (
+				!normalisedAllowedRequestMethods.includes(request.method.toUpperCase())
+			) {
+				response.header('Allow', normalisedAllowedRequestMethods.join(', '));
+				response.status(405).send(options.message);
+				return next(new MethodNotAllowedError(options.message));
```
rowanmanning, line 56: **suggestion:** we can reduce the amount of code we maintain here by falling back to the way Express handles errors. Calling `next` with a new error means you don't need to set the status and send a message. This has the added benefit of error handling middleware (e.g. our logging middleware) being used.

I'd also be tempted to move the check for `response.headersSent` inside here _only_ wrapping the place where we try and set headers. The `next` is fine because Express itself will handle things if headers have already been sent.

```suggestion
				if (!response.headersSent) {
					response.header('Allow', normalisedAllowedRequestMethods.join(', '));
				}
				return next(new MethodNotAllowedError(options.message));
```

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,112 @@
...
+			if (options.logger) {
+				logRecoverableError({
+					error,
+					logger: options.logger,
+					request
+				});
+			}
+			next(error);
+		}
+	};
```
rowanmanning, line 71: **suggestion:** I think we can not bother with this wrapping try/catch because:

1. we're not doing anything async in this function which means there's no risk of crashing the app
2. If we let the error throw instead, then Express will catch it and run the default error handlers anyway, so we get the logging etc for free

rowanmanning (commented):
This is looking great so far, I decided to swoop in with a review on your draft because I think these changes will reduce the number of tests and types you have to write by quite a lot!

packages/middleware-allow-request-methods/lib/index.js
CyntiBinti: Ah okie, thank you for clarifying! :)

packages/middleware-allow-request-methods/lib/index.js
CyntiBinti: Ok cool, thank you for the clarification. This is helpful!

packages/middleware-allow-request-methods/lib/index.js
CyntiBinti: Yeah was wondering about this, but yeah this makes sense as not handling async activities. Cool will remove, thank you!

packages/middleware-allow-request-methods/lib/index.js
CyntiBinti: Nice, ok will bare in mind for next time

packages/middleware-allow-request-methods/README.md
```diff
@@ -0,0 +1,101 @@
...
+// Define your routes
+app.get('/', (req, res) => {
+  res.send('Homepage');
+});
+
+app.post('/submit', (req, res) => {
+  res.send('Form submitted');
+});
+
+app.listen(3000, () => console.log('Server running on port 3000'));
```
rowanmanning, line 53: **suggestion:** the example here doesn't work for a few reasons, I've included some suggestions to make it more comprehensive as well:

  1. The `import` signature change outlined above
  2. The signature for the methods is wrong - they don't accept just an array of strings, they need to be an object with the `allowMethods` option
  3. the `app.use` for `/` will take precedence over the `app.use` for `/submit`. This means that `/submit` allows no request methods (try `GET` and `POST` on submit, both with error). The order would have to change _or_ we should use route-specific middleware
  4. We're missing some docs on global middleware and route-specific middleware that _doesn't_ impact sub-routes

```suggestion
const express = require('express');
const { allowRequestMethods } = require('@dotcom-reliability-kit/middleware-allow-request-methods');

const app = express();

// Allow only certain request methods for the entire app. If you're
// doing this, it must be above ALL routes you want it to apply to:
app.use(allowRequestMethods({ allowedMethods: ['GET', 'HEAD', 'POST'] }));

// Allow only certain request methods for a specific route, e.g. here
// we only allow `GET` and `HEAD` methods for the home page. Note that
// we have to use `all` for the allowed methods here THEN define the get
// request handler:
app
	.route('/')
	.all(allowRequestMethods({ allowedMethods: ['GET', 'HEAD'] }))
	.get((request, response) => {
		response.send('Homepage');
	});

// You can also allow methods for a subset of routes. Remember that this
// applies for all routes that START with the value. E.g. the following
// will also only allow POST requests on `/submit/example`:
app.use('/submit', allowRequestMethods({ allowedMethods: ['POST'] }));
app.post('/submit', (request, response) => {
	response.send('Form submitted');
});

app.listen(3000, () => console.log('Server running on port 3000'));
```

packages/middleware-allow-request-methods/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
```
rowanmanning, line 55: **suggestion:** I'd maybe add some info about what we recommend / the benefits of the above approaches too.

---

We recommend always using this middleware globally with `app.use` as a first middleware in your app. This is because, if a bad actor is making requests to your app to find attack vectors, you throw their request out as early as possible.

Route-specific blocking of methods is an additional layer of protection you can explore. It may be that your app _does_ support `POST` requests for a form but the main view is `GET` only. You can filter out further junk requests on a per-route basis by using the `app.route('...').all()` method or `use` with a path.

packages/middleware-allow-request-methods/types/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@dotcom-reliability-kit/middleware-allow-request-methods' {}
```
rowanmanning, line 1: **suggestion:** we still need types for this. If you don't want to work them out then see the spoiler below 😁

<details><summary>SPOILER - types for this module</summary>

```ts
import { RequestHandler } from 'express';

declare module '@dotcom-reliability-kit/middleware-allow-request-methods' {
	export type RequestMethodOptions = {
		allowedMethods: string[];
	};

	export function allowRequestMethods(
		options: RequestMethodOptions
	): RequestHandler;
}

```

</details>

Once you have types you can simplify the main JS file by removing JSDoc `@typedef` in favour of importing, e.g. the following:

```js
/**
 * @typedef {object} RequestMethodOptions
 * @property {string[]} [allowedMethods] The HTTP methods that are allowed i.e. will not throw 405 errors.
 */
```

becomes:

```js
/**
 * @import { RequestMethodOptions } from '@dotcom-reliability-kit/middleware-allow-request-methods'
 */
```

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+	return function allowRequestMethodsMiddleware(request, response, next) {
+		// If headers are already sent, pass the error to the default Express error handler
+		if (!response.headersSent) {
+			response.header('Allow', normalisedAllowedRequestMethods.join(', '));
+		}
+
+		// If the incoming request method is not in the allowed methods array, then send a 405 error
+		if (
+			!normalisedAllowedRequestMethods.includes(request.method.toUpperCase())
```
rowanmanning, line 47: **suggestion:** this is my bad because my library does this, but `request.method` is _always_ upper-case so we can save a little processing and test edge cases (the `handles case-insensitive method matching` test can be removed too).

We do still need to upper-case the options though because these are untrusted - they're from a user of the library.

```suggestion
			!normalisedAllowedRequestMethods.includes(request.method)
```

packages/middleware-allow-request-methods/test/unit/lib/index.spec.js
```diff
@@ -0,0 +1,130 @@
+const { allowRequestMethods } = require('../../../lib/index');
+const { UserInputError } = require('@dotcom-reliability-kit/errors');
+
+// Mock Express request and response objects
+let mockRequest;
+let mockResponse;
+let mockNext;
+
+describe('allowRequestMethods', () => {
```
rowanmanning, line 9: **praise:** these tests are super comprehensive, really excellent work, thanks!

packages/middleware-allow-request-methods/lib/index.js
CyntiBinti: No all good, thanks as TIL!

packages/middleware-allow-request-methods/README.md
CyntiBinti: Thank you for this clear explanation will add to the README!

packages/middleware-allow-request-methods/README.md
CyntiBinti: Yeah I wasn't too sure as API routing is relatively new for me but this makes sense and thanks for detailing it out 🙌🏽

packages/middleware-allow-request-methods/types/index.d.ts
CyntiBinti: Woohoo tried without looking, using the other middleware index.d.ts file as ref. Didn't realise that's where it was pulling it from! Thanks for this, very helpful :)

packages/middleware-allow-request-methods/types/index.d.ts
```diff
@@ -0,0 +1,11 @@
+import { RequestHandler } from 'express';
+
+declare module '@dotcom-reliability-kit/middleware-allow-request-methods' {
+	export type RequestMethodOptions = {
+		allowedMethods?: string[];
+	};
+
+	declare function allowRequestMethods(
+		options?: RequestMethodOptions
+	): RequestHandler;
```
rowanmanning, line 10: **suggestion:** really small nitpick on this, sorry, but the `allowedMethods` and `options` in these types should probably not be optional. The function errors without them and so we should mark them as required in the types to help hint at issues at build time instead of run time.

```suggestion
	export type RequestMethodOptions = {
		allowedMethods: string[];
	};

	declare function allowRequestMethods(
		options: RequestMethodOptions
	): RequestHandler;
```

packages/middleware-allow-request-methods/lib/index.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+/**
+ * @import { RequestHandler } from 'express'
+ */
+
+/**
+ * Create a middleware function to return 405 (rather than 404) for disallowed request methods.
+ *
+ * @param {RequestMethodOptions} [options]
```
rowanmanning, line 14: **suggestion:** sorry didn't spot this before but along a similar line, `options` here should not be optional

```suggestion
 * @param {RequestMethodOptions} options
```

rowanmanning (commented):
Two really small type-related things and then we're ready to go! Exciting :) thanks for all the back and forth on this

packages/middleware-allow-request-methods/types/index.d.ts
CyntiBinti: Nice good spot thanks!

rowanmanning (approved):
Woo! Thanks, let's ship it 🚀 first new package in a while

PR #1332, opened 2025-02-13
test: wait for logs to appear in end-to-end test

rowanmanning:


packages/middleware-log-errors/test/end-to-end/index.spec.js
```diff
@@ -28,6 +29,18 @@ describe('@dotcom-reliability-kit/middleware-log-errors end-to-end', () => {
...
+
+			// Pino uses async logs so we can't guarantee that our log will have been output
+			// immediately. We wait a total of 5s for the log to come through, checking for
+			// the log in 100ms increments
+			let waitTime = 0;
+			while (waitTime < 5000) {
+				if (stdout.includes('HANDLED_ERROR')) {
+					return;
+				}
+				await setTimeout(100);
```
CyntiBinti, line 41: question: `waitTIme` variable is not incremented so the loop won't exit?

rowanmanning: Haha well spotted, thanks!

PR #1363, opened 2025-03-25
fix: disable log prettification on AWS

rowanmanning:
pino-pretty doesn't seem to work at all when sending logs to AWS CloudWatch. We also never want prettified logs to be in production, so this protects us against the case where `NODE_ENV` isn't set in an AWS context.


CyntiBinti (approved):
THANK YOU 🤩

PR #1369, opened 2025-03-26
test: add end-to-end tests for opentelemetry

rowanmanning:
I wasn't super confident after the recent major update and it led to me doing lots of tests across apps, symlinking in the new package version, and manually checking that metrics and traces were still being sent.

I realised that this could be automated. I've added an end-to-end test where we spin up two Node.js apps - one collector and one regular app with OpenTelemetry added. We capture logs from both and this allows us to verify the basic case of "do metrics/traces get from an app to the configured collector?"

OpenTelemetry uses Protobuf to communicate and it's a real pain to get the parser working. I spent an hour then decided to give up - the key thing is that we see the application's system code in the body that's sent to the collector, we don't want to end up testing OpenTelemetry internals by verifying that the body is valid.


CyntiBinti (approved):
nice 👌🏽

PR #1375, opened 2025-03-28
docs: improve mock logger sample

phelipemaia:
This is just a simple improvement in the docs so the example we use to mock logger also covers cases where the developer invoked Logger constructor in their code.

Feel free to reject if you don't think that's useful or change in case you think there is a better way of doing it.


rowanmanning (approved):
Very nice, thanks for doing this!

PR #1405, opened 2025-04-22
fix: update auto-instrumentations

rowanmanning:
@opentelemetry/auto-instrumentations-node now imports and initialises @opentelemetry/instrumentation-runtime-node by default, as of v0.58.0.

Because of this we no longer need to intialise it or install it ourselves, this is not a breaking change for us because the initialisation is just moving.

### DIFF
--- a/packages/opentelemetry/lib/config/instrumentations.js
+++ b/packages/opentelemetry/lib/config/instrumentations.js
@@ -1,9 +1,6 @@
 const {
 	getNodeAutoInstrumentations
 } = require('@opentelemetry/auto-instrumentations-node');
-const {
-	RuntimeNodeInstrumentation
-} = require('@opentelemetry/instrumentation-runtime-node');
 const { logRecoverableError } = require('@dotcom-reliability-kit/log-error');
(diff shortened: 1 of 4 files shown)


packages/opentelemetry/lib/config/instrumentations.js
```diff
@@ -21,17 +18,14 @@ const IGNORED_REQUEST_PATHS = ['/__gtg', '/__health', '/favicon.ico'];
...
-		new RuntimeNodeInstrumentation()
-	];
+	return getNodeAutoInstrumentations({
+		'@opentelemetry/instrumentation-http': {
+			ignoreIncomingRequestHook
+		},
+		'@opentelemetry/instrumentation-pino': {
+			enabled: false
+		}
+	});
```
rowanmanning, line 28: The `NodeSDKConfiguration['instrumentations']` type allows either an array of arrays of instrumentations (what we were doing) or an array of instrumentations. This means we can safely just switch to returning the result of the function call instead of wrapping in an unnecessary array.

CyntiBinti (approved):

PR #1476, opened 2025-08-20
fix: correct some type issues with Logger

rowanmanning:
These are all issues that raised their head when a TypeScript project in Specialist uses lib checks. Still investigating fully but these are some quick wins to reduce the number of errors


ivomurrell (approved):
yeah maybe I can help look over all the default export mess to see if another pair of eyes can help at some point 😅

